PR #13, opened 2017-10-13
feature/browzine-summon-adapter-rework-BZ-4161

tiandavis:
Refactors to a MutationObserver and jQuery data sniffing to reduce reliance on Summon loading 3rd party scripts effectively. See, https://thirdiron.atlassian.net/browse/BZ-4161

Demo URL:
http://msulibrariestest.summon.serialssolutions.com/#!/search?ho=t&l=en&q=new%20england%20journal%20of%20medicine

Proposed Usage? We ask library developers to use this code snippet in their Summon 2.0 External Script:
```
var browzine = {
  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
  apiKey: "ENTER API KEY"
};

browzine.script = document.createElement("script");
browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzine-summon-adapter.js";
document.head.appendChild(browzine.script);
```

Then we can push out enhancement of all sorts without the library ever having to touch the code again. :) Also, less code for the library to handle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,10 @@ browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzin
 document.head.appendChild(browzine.script);
 ```
 
-Summon 2.0 External Script
+Summon Custom Scripts are added in the Summon 2.0 External Script portion of the Summon Editor Settings:
+![Summon 2.0 Editor Link](https://i.imgur.com/HJFpDGm.png "Summon 2.0 Editor Link")
+
+As a Custom Script Url:
 ![Summon 2.0 External Script](https://i.imgur.com/piLMSic.png "Summon 2.0 External Script")
 
 
--- a/README.md
+++ b/README.md
@@ -1,17 +1,29 @@
 # Library Developers
 
 ## Overview
-Use `browzine-summon-adapter` to enhance Summon search results with BrowZine data; Adds links to Journal and Article content in BrowZine; Uses BrowZine Journal covers.
+Use `browzine-summon-adapter` to enhance Summon search results with BrowZine data; Adds links to Journal and Article content in BrowZine; Uses BrowZine Journal covers. The BrowZine Summon adapter supports IE11+ and evergreen browsers (Chrome, Firefox, Safari, Microsoft Edge).
+
 
 ![Article in Context links in Summon results](https://i.imgur.com/B34LEec.png "Article in Context links in Summon results")
 
 ## How to request your library API endpoint
 Visit Third Iron support to request your library API endpoint - http://support.thirdiron.com/
 
-You will receive your `api` endpoint and your `apiKey`, update the Summon script with these values.
+You will receive your `api` endpoint and your `apiKey`, update the following code snippet with these values and add to your Summon Custom Script:
 
-## What source code to use?
-Use `/src/browzine-summon-adapter.js`, we support IE10+ and evergreen browsers (Chrome, Firefox, Safari, Microsoft Edge).
+```
+var browzine = {
+  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
+  apiKey: "ENTER API KEY"
+};
+
+browzine.script = document.createElement("script");
+browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzine-summon-adapter.js";
+document.head.appendChild(browzine.script);
+```
+
+Summon 2.0 External Script
+![Summon 2.0 External Script](https://i.imgur.com/piLMSic.png "Summon 2.0 External Script")
 
 
 # Contributors
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You will receive your `api` endpoint and your `apiKey`, update the following cod
 ```
 var browzine = {
   api: "https://api.thirdiron.com/public/v1/libraries/XXX",
-  apiKey: "ENTER API KEY"
+  apiKey: "ENTER API KEY",
 };
 
 browzine.script = document.createElement("script");
@@ -28,6 +28,22 @@ Summon Custom Scripts are added in the Summon 2.0 External Script portion of the
 As a Custom Script Url:
 ![Summon 2.0 External Script](https://i.imgur.com/piLMSic.png "Summon 2.0 External Script")
 
+## Customize The BrowZine Enhancement
+
+Customize the naming conventions for each type of search result - Journal/Article - by changing the wording in the quotes below:
+
+E.g. You can customize "View the Journal" and "View Complete Issue" or "Browse Now". These customizations are optional and the defaults are shown below.
+
+```
+var browzine = {
+  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
+  apiKey: "ENTER API KEY",
+  journalWording: "View the Journal",
+  articleWording: "View Complete Issue",
+  journalBrowZineWebLinkText: "Browse Now",
+  articleBrowZineWebLinkText: "Browse Now",
+};
+```
 
 # Contributors
 
@@ -53,7 +69,7 @@ Contributors should:
 - Make the needed changes in `/src/browzine-summon-adapter.js`.
 - Add a test for the change in `/tests/browzine-summon-adapter.js`.
 
-### Deploying (Internal Use Only)
+### Deploying
 
 CircleCI executes deployments to S3 using the deploy-staging and deploy-production npm commands. Neither feature branches nor hotfix branches are deployed to S3.
 
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -149,14 +149,7 @@ browzine.search = (function() {
   };
 
   function getScope(documentSummary) {
-    var token = Object.getOwnPropertyNames(documentSummary).filter(function(property) {
-      property = property.replace(/[^a-z]/gi, "");
-      return property.indexOf("jQuery") === 0;
-    });
-
-    var scope = documentSummary[token].$scope;
-
-    return scope;
+    return angular.element(documentSummary).scope();
   };
 
   function resultsWithBrowZine(documentSummary) {
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -1,9 +1,8 @@
-var browzine = {
-  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
-  apiKey: "ENTER API KEY"
-};
-
 $(function() {
+  if(typeof browzine === "undefined" || browzine === null) {
+    return;
+  }
+
   var api = browzine.api;
   var apiKey = browzine.apiKey;
   var bookIcon = "https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine_open_book_icon.png";
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -194,7 +194,7 @@ browzine.search = (function() {
 }());
 
 $(function() {
-  if(typeof browzine === "undefined" || browzine === null) {
+  if(!browzine) {
     return;
   }
 
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -9,7 +9,8 @@ $(function() {
 
   function isArticle(data) {
     if(typeof data.document !== "undefined" && data.document !== null) {
-      return data.document.content_type.trim() === "Journal Article";
+      var contentType = data.document.content_type.trim();
+      return contentType === "Journal Article";
     } else {
       return data.type === "articles";
     }
@@ -17,7 +18,8 @@ $(function() {
 
   function isJournal(data) {
     if(typeof data.document !== "undefined" && data.document !== null) {
-      return data.document.content_type.trim() === "Journal";
+      var contentType = data.document.content_type.trim();
+      return contentType === "Journal" || contentType === "eJournal";
     } else {
       return data.type === "journals";
     }
@@ -30,10 +32,8 @@ $(function() {
       issn = scope.document.issns[0].trim().replace('-', '');
     }
 
-    if(issn === "") {
-      if(typeof scope.document.eissns !== "undefined" && scope.document.eissns !== null) {
-        issn = scope.document.eissns[0].trim().replace('-', '');
-      }
+    if(typeof scope.document.eissns !== "undefined" && scope.document.eissns !== null && issn === "") {
+      issn = scope.document.eissns[0].trim().replace('-', '');
     }
 
     return encodeURIComponent(issn);
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -135,7 +135,9 @@ $(function() {
     return template;
   };
 
-  function documentSummary(scope, element) {
+  function searchResultsWithBrowZine(documentSummary) {
+    var scope = angular.element(documentSummary).scope();
+
     if(!shouldEnhance(scope)) {
       return;
     }
@@ -149,18 +151,13 @@ $(function() {
 
       if(browzineWebLink) {
         var template = buildTemplate(data, browzineWebLink, bookIcon);
-        $(element).find(".docFooter .row:eq(0)").append(template);
+        $(documentSummary).find(".docFooter .row:eq(0)").append(template);
       }
 
       if(coverImageUrl) {
-        $(element).find(".coverImage img").attr("src", coverImageUrl);
+        $(documentSummary).find(".coverImage img").attr("src", coverImageUrl);
       }
     });
-  };
-
-  function browzineEnhance(element) {
-    var scope = angular.element(element).scope();
-    documentSummary(scope, element);
   };
 
   var results = document.querySelector("#results");
@@ -175,14 +172,14 @@ $(function() {
   var documentSummaries = results.querySelectorAll(".documentSummary");
 
   Array.prototype.forEach.call(documentSummaries, function(documentSummary) {
-    browzineEnhance(documentSummary);
+    searchResultsWithBrowZine(documentSummary);
   });
 
   var observer = new MutationObserver(function(mutations) {
     mutations.forEach(function(mutation) {
       if(mutation.attributeName === "document-summary") {
         var documentSummary = mutation.target;
-        browzineEnhance(documentSummary);
+        searchResultsWithBrowZine(documentSummary);
       }
     });
   });
--- a/src/browzine-summon-adapter.js
+++ b/src/browzine-summon-adapter.js
@@ -1,9 +1,8 @@
-// Begin BrowZine-Summon Integration Code
-angular.module("summonApp.directives")
-.constant("api", "https://api.thirdiron.com/public/v1/libraries/XXX")
-.constant("apiKey", "ENTER API KEY")
-.constant("bookIcon", "https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine_open_book_icon.png")
-.directive("documentSummary", ["$http", "$sce", "api", "apiKey", "bookIcon", function(http, sce, api, apiKey, bookIcon) {
+$(function() {
+  var api = browzine.api;
+  var apiKey = browzine.apiKey;
+  var bookIcon = "https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine_open_book_icon.png";
+
   function isArticle(data) {
     if(typeof data.document !== "undefined" && data.document !== null) {
       return data.document.content_type.trim() === "Journal Article";
@@ -25,6 +24,12 @@ angular.module("summonApp.directives")
 
     if(typeof scope.document.issns !== "undefined" && scope.document.issns !== null) {
       issn = scope.document.issns[0].trim().replace('-', '');
+    }
+
+    if(issn === "") {
+      if(typeof scope.document.eissns !== "undefined" && scope.document.eissns !== null) {
+        issn = scope.document.eissns[0].trim().replace('-', '');
+      }
     }
 
     return encodeURIComponent(issn);
@@ -61,11 +66,11 @@ angular.module("summonApp.directives")
   };
 
   function getData(response) {
-    return Array.isArray(response.data.data) ? response.data.data[0] : response.data.data;
+    return Array.isArray(response.data) ? response.data[0] : response.data;
   };
 
   function getIncludedJournal(response) {
-    return Array.isArray(response.data.included) ? response.data.included[0] : response.data.included;
+    return Array.isArray(response.included) ? response.included[0] : response.included;
   };
 
   function getBrowZineWebLink(data) {
@@ -81,12 +86,14 @@ angular.module("summonApp.directives")
   function getCoverImageUrl(data, response) {
     var coverImageUrl = null;
 
-    if(typeof data.coverImageUrl !== "undefined" && data.coverImageUrl !== null) {
-      coverImageUrl = data.coverImageUrl;
+    if(isJournal(data)) {
+      if(typeof data.coverImageUrl !== "undefined" && data.coverImageUrl !== null) {
+        coverImageUrl = data.coverImageUrl;
+      }
     }
 
     if(isArticle(data)) {
-      if(typeof response.data.included !== "undefined" && response.data.included !== null) {
+      if(typeof response.included !== "undefined" && response.included !== null) {
         var journal = getIncludedJournal(response);
 
         if(typeof journal.coverImageUrl !== "undefined" && journal.coverImageUrl !== null) {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -1,17 +1,29 @@
...
+Use `browzine-summon-adapter` to enhance Summon search results with BrowZine data; Adds links to Journal and Article content in BrowZine; Uses BrowZine Journal covers. The BrowZine Summon adapter supports IE11+ and evergreen browsers (Chrome, Firefox, Safari, Microsoft Edge).
+
 
 ![Article in Context links in Summon results](https://i.imgur.com/B34LEec.png "Article in Context links in Summon results")
 
 ## How to request your library API endpoint
 Visit Third Iron support to request your library API endpoint - http://support.thirdiron.com/
 
-You will receive your `api` endpoint and your `apiKey`, update the Summon script with these values.
+You will receive your `api` endpoint and your `apiKey`, update the following code snippet with these values and add to your Summon Custom Script:
```
karlbecker: 👍 Much better!

src/browzine-summon-adapter.js
```diff
@@ -61,11 +66,11 @@ angular.module("summonApp.directives")
   };
 
   function getData(response) {
-    return Array.isArray(response.data.data) ? response.data.data[0] : response.data.data;
+    return Array.isArray(response.data) ? response.data[0] : response.data;
```
karlbecker: So does `http.get` do something a lot different than `$.getJSON` that we don't have to access things through `.data.data`?  Just curious.

src/browzine-summon-adapter.js
```diff
@@ -124,30 +131,63 @@ angular.module("summonApp.directives")
...
-      if(!shouldEnhance(scope)) {
-        return;
+  function documentSummary(scope, element) {
+    if(!shouldEnhance(scope)) {
+      return;
+    }
+
+    var endpoint = getEndpoint(scope);
+
+    $.getJSON(endpoint, function(response) {
```
karlbecker: Is this less of an "angular-y" way of doing things, or still perfectly valid?  

Does an angular app pretty much always have jQuery installed?

src/browzine-summon-adapter.js
```diff
@@ -124,30 +131,63 @@ angular.module("summonApp.directives")
...
+        $(element).find(".coverImage img").attr("src", coverImageUrl);
+      }
+    });
+  };
 
-      http.get(sce.trustAsResourceUrl(endpoint)).then(function(response) {
-        var data = getData(response);
-        var browzineWebLink = getBrowZineWebLink(data);
-        var coverImageUrl = getCoverImageUrl(data, response);
+  function browZineEnhance(element) {
```
karlbecker: Total nit, but since "browzine" is kind of its own thing, the camel case of `browzineEnhance` might be a little clearer here - otherwise, I read it as "brow zine enhance" in my head 😊

src/browzine-summon-adapter.js
```diff
@@ -124,30 +131,63 @@ angular.module("summonApp.directives")
...
+    var secret = Object.getOwnPropertyNames(element).filter(function(property) {
```
karlbecker: The name `secret` here has me a little confused - this isn't a secret key or anything, correct?  Looks like this is the name of an element's property that is kind of hard to find because of how Angular builds a page?

src/browzine-summon-adapter.js
```diff
@@ -124,30 +131,63 @@ angular.module("summonApp.directives")
...
+    characterData: true,
+    subtree: true,
   };
-}]);
 
-// End BrowZine-Summon Integration Code
+  //Enhance any documentSummary elements present before the observer starts
+  var documentSummaries = results.querySelectorAll(".documentSummary");
+
+  Array.prototype.forEach.call(documentSummaries, function(documentSummary) {
```
karlbecker: Does using `Array.prototype.forEach` mean we have more backward compatibility than if we did `documentSummaries.forEach` ?

tiandavis: Yep. IE and older versions of Safari don't have a `forEach` on `querySelectorAll` results and I try to use jQuery for only DOM manipulation and AJAX convenience methods.

src/browzine-summon-adapter.js
tiandavis: It's for this type of property `jQuery321096145541210543752`, which we only need to access the Angular $scope with all the data for the search result:

image

Someone put in work to make it not easily accessible and I googled and couldn't find what it is exactly, `secret` was nice and short, so it worked for me. Suggestions?

src/browzine-summon-adapter.js
tiandavis: Awww, yeah, I went back-and-forth on that. Thanks!

src/browzine-summon-adapter.js
tiandavis: Yep, Angular is dependent on jQuery, so jQuery is always installed. The whole reason for the rewrite is to decouple from Angular so the BrowZine Summon Adapter runs 100% of the time and we eliminate the risk of Summon never fixing their async LABjs use to load 3rd-Party Angular extensions correctly - i.e. Before starting the manual Angular bootstrap. My point is this update makes this a JavaScript app that tries to read Angular data.

It doesn't run on the Angular framework at this point. We could switch back in the future, of course, but Summon would have to fix that LABjs issue and wait until all scripts are loaded before starting the Angular bootstrap.

src/browzine-summon-adapter.js
tiandavis: Yeah, I was surprised too. Looks like Angular's `http.get` was wrapping our Public API response in a `data` object. Both awkward and surprising it would modify another service's http response. The base jQuery AJAX calls it extends don't do that.

tiandavis:
Hi @karlbecker! Added the MSU Sandbox demo URL to the description. The automated tests were within the Angular framework and because the solution had to move away from depending on Angular to run, the entire testing infrastructure may be unusable. Looking to see what we can do, but not coming up with much luck yet. If anything, will remove temporarily to get this moving along and create a JIRA for the same. Thanks for the review!

karlbecker:
> If anything, will remove temporarily to get this moving along and create a JIRA for the same.

Yeah, seems right - just don't want us to have totally broke tests be fouling up CircleCI.  Better to admit "we can't test this yet" and have the tests disabled... but I think we better get it tested soon after then. 😊

src/browzine-summon-adapter.js
karlbecker: Weird! Nice that we can get rid of that weirdness.

tiandavis:
Yeah, we definitely need to bring this under test. John's rolling it out and I know he probably has expansion plans too. So we'll want to guard against regressions.

src/browzine-summon-adapter.js
karlbecker: Roger that.  Maybe `propertyCreatedForBZIntegration` ?  (If I'm understanding what it's doing correctly, I think that better describes what it is)

src/browzine-summon-adapter.js
karlbecker: Thanks for the clarification!

src/browzine-summon-adapter.js
tiandavis: Was `jQuery321096145541210543752` actually created for BZ integration though? xD

johnseguin:
@KeeganJ --

#1 I surveyed a group of our Summon customers before we did this project and about 50% in my sample used some type of external script.  Summon's interface allows for only ONE Script URL to be used so those who are doing "something" are accustomed to adding various add-ons to this space.

I was discussing on Slack this morning that we might want to consider this tidy version as an option but then ALSO still make our source available for libraries wanting to customize things - for example, maybe they don't want all these features, just some of them (especially as we build the new SKU and there are a lot of options) or they want to localize for a language - we have a large number of customers overseas - particularly in Sweden, Norway and Germany/Switzerland - who might like a local language in there.  Thanks!

tiandavis:
1. Fine. Fine, question! I'll throw something on this PR and have it auto-deploy to s3. That way customers have three integration options. Load our dynamic script direct. Or place our dynamic script at the bottom of their existing custom script. Plus what @johnseguin said! I'll also get those customizable features turned into properties, options are good.

2. With this setup we'll be able to do whatever we want during deployment. We ran into some issues with Babel inserting "use strict", which broke those with existing custom scripts that don't conform. Had a tough time getting Babel to stop injecting that piece and template liberals broke the compiler with new lines (which is the whole point of template literals), so stood down for the time being, but we can definitely revisit. If we can't stop the use script, we may be able to wrap the output in a closure so as to not affect code that's not ours, just not sure.

src/browzine-summon-adapter.js
karlbecker: Aha, then I do not understand correctly what it's doing 😊 

I thought it was being created as a result of the BZ summon instance and was sitting somewhere in the HTML, but I'm not seeing that looking at the page now.  

`secret` or `token` is fine then, don't mind me, I just didn't fully understand what was going on. 😁 

Still kind of curious where in the code this is appearing, since I don't see it in http://msulibrariestest.summon.serialssolutions.com/#!/search?ho=t&l=en&q=new%20england%20journal%20of%20medicine , but no big deal - don't let my curiosity here hold us up from merging. 👍

tiandavis:
Awww, now I remember why we can't host any "entire" piece of the process for them @KeeganJ. It's because of credentials and the fact that they only have ONE script URL at their disposal. So it's not like they can host their credentials in one and link to our code snippet in the other. It's pretty much they host their credentials + our dynamic script snippet; OR they host their credentials and our ENTIRE src code. 

Fantastic question though @KeeganJ! Thanks!

src/browzine-summon-adapter.js
tiandavis: Thanks Karl! `jQuery321096145541210543752 ` seems to be buried in the jQuery/Angular internals somehow. Summon is basically getting data from their database for the search result and then storing it on Angular's $scope object. We're definitely not supposed to be using it, feels kinda naughty, hehehe.

I was only able to find it by doing a `$('.documentSummary')` and then inspecting the jQuery objects in the results. No crystal ball here, but I know we'll need to collaborate w/ the Summon team about the LABjs issue at some point, I'm just not sure when strategy-wise TBH. Some good brandy or whiskey might help shake out some thoughts though. :)

karlbecker: Google says Serials Solutions HQ is in Seattle... from my little experience there, a craft beer, PBR, or [tiki drink](https://www.yelp.com/biz/lava-lounge-seattle?osq=tiki+bar) could work. 🍻  🍸

tiandavis:
Quick Update: 

Found an official way to check for the scope over parsing the `jQuery321096145541210543752` object: https://github.com/thirdiron/browzine-summon-adapter/pull/13/files#diff-f742f80a6fce9a21cc8759359a0a2464R158

SPOILER ALERT: Not looking like we'll be able to bring this under test until Summon fixes it's LABjs async issue and we're able to revert back to implementing this in an Angular context. Threw the kitchen sink at bringing this under test w/ the MutationObserver refactor, but running into async timing issue, making the individual test runs nondeterministic.

@johnseguin About to stub out the customization properties now and we should be able to wrap this up!

Thanks Folks!

karlbecker:
> SPOILER ALERT: Not looking like we'll be able to bring this under test until Summon fixes it's LABjs async issue and we're able to revert back to implementing this in an Angular context. 

Bummer! 

> Found an official way to check for the scope over parsing the jQuery321096145541210543752 object

Nice!  Much cleaner than before, I dig it 👍 

Aside from that, only thought about the new changes:

Instead of commenting out the content of the `it`s, can we do something like the `it.skip` that mocha has?  

Thanks @tiandavis !

tiandavis:
> Instead of commenting out the content of the its, can we do something like the it.skip that mocha has?
We can using `xdescribe` or `xit`, but then no tests will run and we're back to the commits failing.

karlbecker:
Okay, never mind then - commented out works.

src/browzine-summon-adapter.js
```diff
@@ -1,9 +1,8 @@
-var browzine = {
-  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
-  apiKey: "ENTER API KEY"
-};
-
 $(function() {
+  if(typeof browzine === "undefined" || browzine === null) {
```
karlbecker: 👍

src/browzine-summon-adapter.js
```diff
@@ -57,21 +81,21 @@ angular.module("summonApp.directives")
   };
 
   function shouldEnhance(scope) {
-    return (isJournal(scope) && getIssn(scope)) || (isArticle(scope) && getDoi(scope));
+    return !!((isJournal(scope) && getIssn(scope)) || (isArticle(scope) && getDoi(scope)));
```
KeeganJ: Is the `!!` necessary here? I think the `&&`s and `||`s will case it into a boolean before it gets to this operator.

src/browzine-summon-adapter.js
```diff
@@ -98,56 +124,103 @@ angular.module("summonApp.directives")
...
     if(isJournal(data)) {
-      assetClass = "View the Journal";
+      wording = browzine.journalWording || "View the Journal";
+      browzineWebLinkText = browzine.journalBrowZineWebLinkText || "Browse Now";
     }
 
     if(isArticle(data)) {
-      assetClass = "View Complete Issue";
+      wording = browzine.articleWording || "View Complete Issue";
+      browzineWebLinkText = browzine.articleBrowZineWebLinkText || "Browse Now";
```
KeeganJ: If this statement is executed regardless of whether the object is a journal or article, we may want to take it out of the block.

src/browzine-summon-adapter.js
```diff
@@ -98,56 +124,103 @@ angular.module("summonApp.directives")
...
     }
 
-    // You can change the underlined "Browse Now" link name on line 122 below.
     var template = "<div class='browzine'>" +
-                     "{{assetClass}}: <a class='browzine-web-link' href='{{browzineWebLink}}' target='_blank' style='text-decoration: underline; color: #333;'>Browse Now</a> " +
+                     "{{wording}}: <a class='browzine-web-link' href='{{browzineWebLink}}' target='_blank' style='text-decoration: underline; color: #333;'>{{browzineWebLinkText}}</a> " +
```
KeeganJ: Worth tossing out that you can do ES5 friendly multiline string by adding a `\` to the end of each line, though the [Airbnb style guide](http://airbnb.io/javascript/#strings--line-length) doesn't seem to like multiline strings at all.

src/browzine-summon-adapter.js
```diff
@@ -98,56 +124,103 @@ angular.module("summonApp.directives")
...
+  return {
+    resultsWithBrowZine: resultsWithBrowZine,
+    getScope: getScope,
+    shouldEnhance: shouldEnhance,
+    getEndpoint: getEndpoint,
+    getData: getData,
+    getBrowZineWebLink: getBrowZineWebLink,
+    getCoverImageUrl: getCoverImageUrl,
+    buildTemplate: buildTemplate,
+  };
```
KeeganJ: This is a much cleaner and more readable end return statement!

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -33,15 +33,17 @@ describe("BrowZine Summon Adapter", function() {
...
 
-      documentSummaryElement = getCompiledDocumentSummaryDirective();
+      search.resultsWithBrowZine(documentSummary);
     });
 
-    it("should have an enhanced browse journal in browzine option", function() {
-      httpBackend.flush();
+    afterEach(function() {
+
+    });
```
KeeganJ: I think this is good to go 😉

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -50,29 +52,31 @@ describe("BrowZine Summon Adapter", function() {
     });
 
     it("should have an enhanced browzine journal cover", function() {
-      httpBackend.flush();
```
KeeganJ: Why aren't we flushing between tests anymore?

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -96,15 +100,17 @@ describe("BrowZine Summon Adapter", function() {
...
 
-      documentSummaryElement = getCompiledDocumentSummaryDirective();
+      search.resultsWithBrowZine(documentSummary);
     });
 
-    it("should have an enhanced browse article in browzine option", function() {
-      httpBackend.flush();
+    afterEach(function() {
+
+    });
```
KeeganJ: Ditto

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -158,15 +166,17 @@ describe("BrowZine Summon Adapter", function() {
...
-      documentSummaryElement = getCompiledDocumentSummaryDirective();
+      search.resultsWithBrowZine(documentSummary);
     });
 
-    it("should not have an enhanced browse article in browzine option", function() {
-      httpBackend.flush();
+    afterEach(function() {
 
-      var template = documentSummaryElement.find(".browzine");
+    });
```
KeeganJ: Ditto, I'll stop leaving notes on these.

tests/unit/models/search.js
```diff
@@ -0,0 +1,287 @@
...
+    expect(search).toBeDefined();
+  });
+
+  it("should not enhance a search result without a document", function() {
+    var scope = {};
+
+    expect(search.shouldEnhance(scope)).toEqual(false);
+  });
+
+  it("should detect whether a journal search result can be enhanced", function() {
```
KeeganJ: Just a wording nitpick here, but since we're only testing the positive case in this test and the two below, could we rename the `it` message to something that conveys that? Such as:

`it("should choose to enhance a journal article with an ISSN")`

tests/unit/models/search.js
```diff
@@ -0,0 +1,287 @@
...
+      }
+    };
+
+    expect(search.getEndpoint(scope)).toContain("articles/doi/10.1136%2Fbmj.h2575?include=journal");
+  });
+
+  it("should include a browzineWebLink in the BrowZine API response for a journal", function() {
+    var data = search.getData(journalResponse);
+    expect(data).toBeDefined();
+    expect(search.getBrowZineWebLink(data)).toEqual("https://browzine.com/libraries/XXX/journals/10292");
```
KeeganJ: I'd move strings like these that are shared with strings set earlier into test constants.

tests/unit/models/search.js
```diff
@@ -0,0 +1,287 @@
...
+    expect(data).toBeDefined();
+    expect(search.getCoverImageUrl(data, journalResponse)).toBeNull();
+  });
+
+  it("should not build a coverImageUrl when the BrowZine API response is missing the article data type", function() {
+    var data = search.getData(articleResponse);
+    delete data.type;
+    expect(data).toBeDefined();
+    expect(search.getCoverImageUrl(data, journalResponse)).toBeNull();
+  });
```
KeeganJ: Excellent job getting all the setup logic into a shared `beforeEach`, it makes the rest of these tests very short and readable and the use cases very apparent

src/browzine-summon-adapter.js
tiandavis: My exact thoughts too until I unit tested it and an issn returned, Eeek...! 🎃
https://github.com/thirdiron/browzine-summon-adapter/pull/13/files#diff-18e41204f346c94301e808d824658669R71

src/browzine-summon-adapter.js
tiandavis: Should be separate customization properties `browzine.articleBrowZineWebLinkText` and `browzine.journalBrowZineWebLinkText`. Figured there could be a library that may want a different message for each scenario.

src/browzine-summon-adapter.js
tiandavis: Yup, tested that out as well. IIRC, there was some issue during testing or maybe IE, can't remember the details, but I think some `\n` newline characters were being injected or something. Settled on 'ol faithful here, hehe.

src/browzine-summon-adapter.js
tiandavis: Yep! Almost went w/ Rebbeca Murphy's [Object Literal Pattern](http://rmurphey.com/blog/2009/10/15/using-objects-to-organize-your-code), but went w/ the Revealing Module Pattern here for it's ability to have both private and public properties/methods. Combining that w/ the `browzine` namespace really helped to expose units of code to test while not polluting the global namespace. It's amazing how code bases evolve.

src/browzine-summon-adapter.js
KeeganJ: 2spooky4me

tests/acceptance/browzine-summon-adapter-test.js
tiandavis: We don't have access to `httpBackend` for now. You get access to the `httpBackend` service when you make AJAX calls in an Angular context (e.g. directive, service, etc.). But we're not able to run in an Angular context until the Summon team loads all scripts before manually bootstrapping angular. Right now in prod, Summon mostly boots before our script loads and when that happens, Angular doesn't know to run our code.

So instead of using `httpBackend` to mock our AJAX calls, I was able to override `$.getJSON` to mock what our API would return. See, https://github.com/thirdiron/browzine-summon-adapter/pull/13/files#diff-a1badd2f04e0cf74e7b88220c9909758R23

tests/unit/models/search.js
tiandavis: Wrestled with this too, but prioritized test independence over reuse.

src/browzine-summon-adapter.js
```diff
@@ -98,56 +124,103 @@ angular.module("summonApp.directives")
...
+}());
+
+$(function() {
+  if(typeof browzine === "undefined" || browzine === null) {
```
mike-lang: Is this meaningfully different from `if(!browzine) {`?

src/browzine-summon-adapter.js
```diff
@@ -98,56 +124,103 @@ angular.module("summonApp.directives")
...
 
-// End BrowZine-Summon Integration Code
+  //Enhance any documentSummary elements present before the observer starts
+  var documentSummaries = results.querySelectorAll(".documentSummary");
+
+  Array.prototype.forEach.call(documentSummaries, function(documentSummary) {
+    browzine.search.resultsWithBrowZine(documentSummary);
+  });
+
+  var observer = new MutationObserver(function(mutations) {
```
mike-lang: Neat. I'd not come across MutationObserver before. TIL

tests/unit/models/search.js
```diff
@@ -0,0 +1,287 @@
...
+    var scope = {
+      document: {
+        dois: ["10.1136/bmj.h2575"]
+      }
+    };
+
+    expect(search.shouldEnhance(scope)).toEqual(false);
+  });
+
+  it("should build a journal endpoint for a journal search result", function() {
```
mike-lang: In general, grouping the tests by method they exercise might make this easier to quickly navigate

src/browzine-summon-adapter.js
tiandavis: Awww, good catch! Looks like `!browzine` checks [more cases](https://stackoverflow.com/a/5941649):
```
false
null
undefined
""
0
NaN
```

Whereas the above only checks or `undefined` and `null`. Thanks!

src/browzine-summon-adapter.js
tiandavis: Yep, it saved our bacon, hehe. Thankfully, we get IE11+ with it too.

tests/unit/models/search.js
tiandavis: This should help w/ organization, I like it. Thanks!

README.md
```diff
@@ -1,18 +1,49 @@
...
+
+```
+var browzine = {
+  api: "https://api.thirdiron.com/public/v1/libraries/XXX",
+  apiKey: "ENTER API KEY",
+};
+
+browzine.script = document.createElement("script");
+browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzine-summon-adapter.js";
+document.head.appendChild(browzine.script);
```
karlbecker: I'm a big fan of this and all the good side-effects that come with it 👍

src/browzine-summon-adapter.js
```diff
@@ -1,30 +1,52 @@
...
-    if(typeof data.document !== "undefined" && data.document !== null) {
-      return data.document.content_type.trim() === "Journal";
-    } else {
-      return data.type === "journals";
+    var result = false;
+
+    if(data.document) {
+      if(data.document.content_type) {
+        var contentType = data.document.content_type.trim();
+        result = (contentType === "Journal" || contentType === "eJournal");
```
karlbecker: This kind of matching and hunting for certain element names we might not know feels a lot like some of the logic we need to do in Crawl Steps in order to find certain elements on a page to click 😊 

One thought: we are doing case-sensitive comparison here, but could we do a case insensitive compare?

Perhaps even better, but more speculative: what if we just look for `journal` to appear, case-insensitively, anywhere in `contentType`?  I'm just thinking if other libraries might call the content type "electronicJournal" or "e-Journal" or something like that... but I don't know!

src/browzine-summon-adapter.js
```diff
@@ -1,30 +1,52 @@
...
+      }
+    }
+
+    if(data.type) {
+      result = (data.type === "journals");
```
karlbecker: Same thing here - what if we just looked if `journal` appeared in `data.type` ?

Of course, maybe that's just getting things to be too loose... not sure, just trying to reduce the amount of future changes we may have to do. But if some other library has a type of, say, "ejournals", eh, it's pretty easy to change our logic, so maybe no change should be done here. Just thinking out loud 😊

src/browzine-summon-adapter.js
```diff
@@ -1,30 +1,52 @@
...
     var issn = "";
 
-    if(typeof scope.document.issns !== "undefined" && scope.document.issns !== null) {
-      issn = scope.document.issns[0].trim().replace('-', '');
+    if(scope.document) {
+      if(scope.document.issns) {
+        issn = scope.document.issns[0].trim().replace('-', '');
+      }
+
+      if(scope.document.eissns && issn === "") {
```
karlbecker: Total nit here, but do we want to use `''` here and on line 40, since we also use single quotes in the `replace` calls near here?

Looks like we're using a mix of double and single quotes throughout the repo - any preference on one or the other?

src/browzine-summon-adapter.js
karlbecker: If we had the ES6 features node version 6 has, or babel provides on Ember, we could even have the wildly clean
```
return {
  resultsWithBrowZine,
  shouldEnhance,
  getEndpoint,
  getData,
  getBrowZineWebLink,
  getCoverImageUrl,
  buildTemplate,
};
```

Just thought I'd mention that in case anyone had not seen this before - it's a nice little cleanup, though doubt we can do it and keep compatible with all the browsers we currently work with.

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -9,15 +9,15 @@ describe("BrowZine Summon Adapter", function() {
 
       documentSummary = $("<div class='documentSummary' document-summary><div class='coverImage'><img src=''/></div><div class='docFooter'><div class='row'></div></div></div>");
 
-      $.extend(documentSummary, {
-        "jQuery321088802690231186742": {
```
karlbecker: 👋 bye bye

README.md
tiandavis: Yup! Along w/ adding tests, we may want to consider doing a similar setup for Primo. It'll be as easy as adding the s3 deployment process we have for Summon and wowzer would enhancements be so clean. +1

src/browzine-summon-adapter.js
tiandavis: Good catch. You should see double quotes mainly, then single quotes when specifying template properties:
```
"<div class='browzine'>"
```

src/browzine-summon-adapter.js
tiandavis: Had the same thought about checking for "journal" when we first started, then ran into this beauty while enhancing article results:
image

"Journal Article". . . 😳  - Ok, fair play Summon, fair play. 😂 

No bueno there, but really like the case-insensitive, will add a test and update the same. Pray to the Gods these Summon types aren't user configurable. 😂

tests/acceptance/browzine-summon-adapter-test.js
tiandavis: 😂

src/browzine-summon-adapter.js
tiandavis: Yeah, the key is getting as much customers as possible using the dynamic script, then enhancements and bug fixes are as simple as a pull request and boom they're productive again.

tiandavis:
Weird, no reply box on the es6 comment. But, Yep, skies the limit when we combine the dynamic script w/ customization variables. Until then...IE ;( 

Look how beautiful this looks. :)
```
return {
  resultsWithBrowZine,
  shouldEnhance,
  getEndpoint,
  getData,
  getBrowZineWebLink,
  getCoverImageUrl,
  buildTemplate,
};
```

src/browzine-summon-adapter.js
karlbecker: Oh well look at Summon pulling that kind of thing 😊 

> Pray to the Gods these Summon types aren't user configurable. 😂

Yeah, do we know?  Does @johnseguin or you have a link to any documentation that might mention whether they are or not?

tiandavis:
Looks fantastic folks! Thanks for all the PR Reviews! 🎉

johnseguin:
@tiandavis Are you waiting on me for something here?

PR #17, opened 2018-02-07
feature/browzine-360-core-adapter-no-api-results-checks-02072018

tiandavis:
## Summary - [BZ-4334](https://thirdiron.atlassian.net/browse/BZ-4334)
Handles a valid ISSN search with no Public API search results. See, [Slack Conversation with John](https://thirdiron.slack.com/archives/C6ND59F7C/p1518022587000057)

## Deploy Precautions
- None

### DIFF
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -95,8 +95,10 @@ browzine.serSol360Core = (function() {
   function getBrowZineWebLink(data) {
     var browzineWebLink = null;
 
-    if(data.browzineWebLink) {
-      browzineWebLink = data.browzineWebLink;
+    if(data) {
+      if(data.browzineWebLink) {
+        browzineWebLink = data.browzineWebLink;
+      }
     }
 
     return browzineWebLink;
@@ -105,11 +107,25 @@ browzine.serSol360Core = (function() {
   function getCoverImageUrl(data) {
     var coverImageUrl = null;
 
-    if(data.coverImageUrl) {
-      coverImageUrl = data.coverImageUrl;
+    if(data) {
+      if(data.coverImageUrl) {
+        coverImageUrl = data.coverImageUrl;
+      }
     }
 
     return coverImageUrl;
+  };
+
+  function getBrowZineEnabled(data) {
+    var browzineEnabled = null;
+
+    if(data) {
+      if(data.browzineEnabled) {
+        browzineEnabled = data.browzineEnabled;
+      }
+    }
+
+    return browzineEnabled;
   };
 
   function buildTemplate(browzineWebLink) {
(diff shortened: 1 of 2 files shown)


src/360-core/browzine-360-core-adapter.js
```diff
@@ -105,13 +107,27 @@ browzine.serSol360Core = (function() {
   function getCoverImageUrl(data) {
     var coverImageUrl = null;
 
-    if(data.coverImageUrl) {
-      coverImageUrl = data.coverImageUrl;
+    if(data) {
```
karlbecker: Small nit, but could we do 
```
if (data && data.coverImageUrl) {
...
```
to save one level of nested `if`?

tests/unit/models/360-core.js
```diff
@@ -149,6 +149,14 @@ describe("SerSol 360 Core Model >", function() {
       expect(data).toBeDefined();
       expect(serSol360Core.getBrowZineWebLink(data)).toEqual("https://browzine.com/libraries/XXX/journals/10292");
     });
+
+    it("should return null when the BrowZine API response for a journal is empty", function() {
+      var data = serSol360Core.getData({
+        "data": []
```
karlbecker: These tests make my heart warm with confidence that we won't regress on handling these cases 😊 ❤️

tests/unit/models/360-core.js
```diff
@@ -165,6 +181,14 @@ describe("SerSol 360 Core Model >", function() {
       expect(data).toBeDefined();
       expect(serSol360Core.getBrowZineEnabled(data)).toEqual(true);
     });
+
+    it("should return null when the BrowZine API response for a journal is empty", function() {
+      var data = serSol360Core.getData({
+        "data": []
+      });
+      expect(data).toBeUndefined();
+      expect(serSol360Core.getBrowZineEnabled(data)).toEqual(null);
```
karlbecker: _non-blocking PR comment_

I'm curious if we'd think `getBrowZineEnabled()` should always return a boolean, or if `null` is desired for some reason, maybe an Angular-specific reason.  

If there's no data, seems like `getBrowZineEnabled()` returning false would be fine, since as far as we know it's not enabled... but perhaps it's even more appropriate to send `null`, because it's neither `true` nor `false`, we just simply do not know since we got no data.

I think this comment is mostly me thinking out loud about what `null` signifies, so I think no change needed here 😊

src/360-core/browzine-360-core-adapter.js
tiandavis: Good one.

tests/unit/models/360-core.js
tiandavis: It's a beautiful thing, indeed.

tests/unit/models/360-core.js
tiandavis: > perhaps it's even more appropriate to send null, because it's neither true nor false, we just simply do not know since we got no data

This is my thought, couldn't have said it better.

karlbecker:
Looks good Tian!

PR #25, opened 2018-03-26
feature/use-identifiers-issn-BZ-4549

tiandavis:
Gets ISSN from title identifiers object instead of extracting from a third-party service.

### DIFF
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -6,31 +6,30 @@ browzine.serSol360Core = (function() {
     return url.replace("api.thirdiron.com", "public-api.thirdiron.com");
   };
 
-  function getQueryVariable(url, key) {
-    var query = url.split("?")[1];
-    var parameters = query.split("&");
-    var value = "";
+  function getIssn(title) {
+    var issn = "", eissn = "";
 
-    for(var i = 0; i < parameters.length; i++) {
-      var pair = parameters[i].split("=");
+    if(title.identifiers) {
+      var issnIdentifier = title.identifiers.filter(function(identifier) {
+        if(identifier.type && identifier.value) {
+          return identifier.type.toLowerCase() === "issn" && identifier.value.length > 0;
+        }
+      }).pop();
 
-      if(pair[0] == key) {
-        value = pair[1];
-        return value;
+      var eissnIdentifier = title.identifiers.filter(function(identifier) {
+        if(identifier.type && identifier.value) {
+          return identifier.type.toLowerCase() === "eissn" && identifier.value.length > 0;
+        }
+      }).pop();
+
+      if(issnIdentifier) {
+        issn = issnIdentifier.value;
+      } else if(eissnIdentifier) {
+        eissn = eissnIdentifier.value;
       }
     }
 
-    return value;
-  };
-
-  function getIssn(title) {
-    var issn = "";
-
-    if(title.syndeticsImageUrl) {
-      issn = getQueryVariable(title.syndeticsImageUrl, "issn");
-    }
-
-    return encodeURIComponent(issn);
+    return encodeURIComponent(issn || eissn);
   };
 
   function getTarget(issn) {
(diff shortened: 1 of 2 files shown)


src/360-core/browzine-360-core-adapter.js
```diff
@@ -6,31 +6,30 @@ browzine.serSol360Core = (function() {
...
+    var issn = "", eissn = "";
 
-    return value;
-  };
+    if(title.identifiers) {
+      var issnIdentifier = title.identifiers.filter(function(identifier) {
+        if(identifier.type && identifier.value) {
+          return identifier.type.toLowerCase() === "issn" && identifier.value.length > 0;
+        }
+      }).pop();
```
KeeganJ: If we find an ISSN here, can we safely short circuit the rest of this method and return it?

tiandavis: Good question Keegan. Doesn't look like there is a built-in ability to break in `filter`. Should be 1 to 2 objects in that `identifiers` array max. Should be fine.

tiandavis: About to push an update, think we may be able to short circuit it.

KeeganJ: Nice!

src/360-core/browzine-360-core-adapter.js
```diff
@@ -6,31 +11,28 @@ browzine.serSol360Core = (function() {
...
+        if(identifier.type && identifier.value) {
+          return identifier.type.toLowerCase() === "eissn" && identifier.value.length > 0;
+        }
+      }).pop();
 
-    if(title.syndeticsImageUrl) {
-      issn = getQueryVariable(title.syndeticsImageUrl, "issn");
+      if(eissnIdentifier) {
+        return encodeURIComponent(eissnIdentifier.value);
+      }
```
KeeganJ: If we don't have an EISSN, we'll end up returning `undefined` here, is that alright? Or do we not expect to get in that case?

tiandavis: Without an EISSN, we won't hit that return block at all. Added some more test cases to validate the same. Good question, Thanks!

tests/unit/models/360-core.js
```diff
@@ -96,17 +121,29 @@ describe("SerSol 360 Core Model >", function() {
...
+      expect(serSol360Core.getIssn(title)).toEqual("2960-1462");
     });
 
-    it("should return empty string when no issn is available on syndeticsImageUrl", function() {
-      var syndeticsImageUrl = scope.searchResultsCtrl.titleData.titles[2].syndeticsImageUrl;
-      var issn = serSol360Core.getQueryVariable(syndeticsImageUrl, "issn");
-      expect(issn).toEqual("");
+    it("should return an empty string if no issn or eissn is available", function() {
+      var titles = serSol360Core.getTitles(scope);
+      expect(serSol360Core.getIssn(titles[5])).toEqual("");
```
karlbecker: Do you think we could improve the readability here a little by assigning `titles[5]` to a named variable that describes what it is?  Like
```
var journalWithEmptyStringForEissn = titles[5];
expect(serSol360Core.getIssn(journalWithEmptyStringForEissn)).toEqual("");
```

Then we can clearly see I the test what the difference is between that line, and the line with `titles[6]`, which could maybe be something like:
```
var journalWithNeitherEissnNorIssn = titles[6];
expect(serSol360Core.getIssn(journalWithNeitherEissnNorIssn)).toEqual("");
```

tests/unit/models/360-core.js
```diff
@@ -127,9 +164,11 @@ describe("SerSol 360 Core Model >", function() {
     });
 
     it("should not enhance a journal search result without an issn or eissn", function() {
-      expect(titles.length).toEqual(2);
+      expect(titles.length).toEqual(4);
       expect(titles[0].title).toEqual("The New England journal of medicine");
       expect(titles[1].title).toEqual("The New England journal of medicine and surgery");
+      expect(titles[2].title).toEqual("Cell");
+      expect(titles[3].title).toEqual("Art");
```
karlbecker: How do we confirm a journal has _not_ been enhanced here? Should we be `expect`ing a property to not exist?

tiandavis: Un-enhanced Journals won't have the `shouldEnhance` property and won't be added to the list of Journals to enhance. Added a test for the same.

tests/unit/models/360-core.js
tiandavis: Realized this probably indicated we needed separate test cases. Ended up splitting out the tests and adding on for the case where there is an empty value for an ISSN identifier.

PR #28, opened 2018-04-13
feature/primo-increase-compatibility-BZ-4546

tiandavis:
## Summary - [BZ-4546](https://thirdiron.atlassian.net/browse/BZ-4546)

Primo hybrid approach where customers use our exposed dynamic services like an SDK and pass the controller `$scope` data, while maintaining any Primo controllers on their side.

## Deploy Precautions

- None

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,28 @@
+## Summary - [BZ-XXXX](https://thirdiron.atlassian.net/browse/BZ-XXXX)
+
+<!--Required: The high level goal of the desired outcome of this PR. This will be included in the auto-generated release notes for a prod deployment.-->
+
+
+
+## Description
+
+<!-- Optional: For going into further detail on the change. Screenshots, gifs, review guidance, etc-->
+
+
+
+## Implementation Notes
+
+<!-- Optional: Any file / API changes done to accomplish the larger goal laid out in the summary.-->
+
+1. ​
+
+## Deploy Precautions
+
+<!-- Required: Potential things to look out for during the deployment process. Delete any of the hazards below that do not apply to this PR. Feel free to go into further detail on any of the points you decide to keep.-->
+
+- :warning: Depends on another PR getting into production first: 
+- :warning: Another PR is waiting on this one to get merged:
+- :warning: Has to be deployed concurrently with another system: 
+- :warning: Potential to break other untested functionality in production:
+- :warning: Other: 
+- None
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ var browzine = {
 };
 
 browzine.script = document.createElement("script");
-browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzine-summon-adapter.js";
+browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/summon/browzine-summon-adapter.js" + Date.now();
 document.head.appendChild(browzine.script);
 ```
 
@@ -84,7 +84,7 @@ var browzine = {
 };
 
 browzine.script = document.createElement("script");
-browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/360-core/browzine-360-core-adapter.js";
+browzine.script.src = "https://s3.amazonaws.com/browzine-adapters/360-core/browzine-360-core-adapter.js" + Date.now();
 document.head.appendChild(browzine.script);
 ```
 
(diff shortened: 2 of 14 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
+## Summary - [BZ-XXXX](https://thirdiron.atlassian.net/browse/BZ-XXXX)
```
KeeganJ: 🙌

README.md
```diff
@@ -114,14 +114,33 @@ var browzine = {
 ### BrowZine Primo Adapter Script
 Update the following code snippet with the `api` endpoint and `apiKey` values:
 ```
```
KeeganJ: FYI, most markdown renderers (including Github's) can use language specific syntax highlighting on code segments if you specify the language on the header like this:

````
```js
console.log('Highlighting!');
```
````

Renders to:
```js
console.log('Highlighting!');
```

It doesn't work for all languages, but if you happen to know the language code that works with your target it's a nice readability add.

Also, apparently you can escape triple backticks by using quadruple backticks above and below so you can display example markdown. TIL.

src/primo/browzine-primo-adapter.js
```diff
@@ -1,78 +1,290 @@
...
+
+    var template = "<div class='browzine'>" +
+                      "<a class='browzine-web-link' href='{browzineWebLink}' target='_blank' title='{browzineWebLinkText} in BrowZine'>" +
+                          "<img src='{bookIcon}' class='browzine-book-icon'/> {browzineWebLinkText}" +
+                      "</a>" +
+                   "</div>";
+
+    template = template.replace(/{browzineWebLink}/g, browzineWebLink);
+    template = template.replace(/{browzineWebLinkText}/g, browzineWebLinkText);
+    template = template.replace(/{bookIcon}/g, bookIcon);
```
KeeganJ: Clever ES5 workaround, make your own template strings!

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -0,0 +1,190 @@
...
+      jasmine.Ajax.uninstall();
+    });
+
+    it("should have an enhanced browse article in browzine option", function() {
+      var template = searchResult.find(".browzine");
+      expect(template).toBeDefined();
+      expect(template.text().trim()).toEqual("View Issue Contents");
+      expect(template.find("a.browzine-web-link").attr("href")).toEqual("https://browzine.com/libraries/XXX/journals/18126/issues/7764583?showArticleInContext=doi:10.1136/bmj.h2575");
+      expect(template.find("a.browzine-web-link").attr("target")).toEqual("_blank");
+      expect(template.find("img.browzine-book-icon").attr("src")).toEqual("https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine_open_book_icon.png");
```
karlbecker: Non-blocking comment: we could use assets.thirdiron.com here if we wanted - it’s an alias to the thirdiron-assets bucket. Not a big deal though.

tests/unit/models/primo.js
```diff
@@ -0,0 +1,680 @@
...
+      var url = "https://public-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(primo.urlRewrite(url)).toEqual("https://public-api.thirdiron.com/public/v1/libraries/XXX");
+    });
+  });
+
+  describe("primo model shouldEnhance method >", function() {
+    it("should not enhance a search result without scope data", function() {
+      var scope = {};
+
+      expect(primo.shouldEnhance(scope)).toEqual(false);
```
karlbecker: Instead of just checking if this returns false, could we pass in the invalid data and confirm it hasn’t messed with it?

src/primo/browzine-primo-adapter.js
```diff
@@ -1,78 +1,290 @@
...
+    var doi = "";
+    var result = getResult(scope);
+
+    if(result && result.pnx) {
+      if(result.pnx.addata && result.pnx.addata.doi) {
+        if(result.pnx.addata.doi[0]) {
+          doi = result.pnx.addata.doi[0].trim().replace("-", "");
+        }
+      }
+    }
```
KeeganJ: This whole file is really well decomposed into small functions!

tests/unit/models/primo.js
```diff
@@ -0,0 +1,680 @@
...
+      var scope = {
+        result: {
+          pnx: {
+            display: {
+              type: ["journal"]
+            },
+
+            addata: {
+              issn: ["0028-4793"],
+              eissn: ["0096-6762"]
```
karlbecker: What if we add a test case where one of these values matches our data, but one does not? (simulating the instance where they have an ISSN or eISSN that we simply don’t have)

We should still enhance it, just want to make sure it enhances it when only one of the many provided match.

README.md
karlbecker: TIL!

tests/unit/models/primo.js
tiandavis: Happy to say we already do! That pretty much describes the journal result acceptance test in a nutshell. Specifically, see lines:
https://github.com/thirdiron/browzine-summon-adapter/pull/28/files#diff-2b01e2be49b9c65a2479fe3472439f6fR67 

https://github.com/thirdiron/browzine-summon-adapter/pull/28/files#diff-2b01e2be49b9c65a2479fe3472439f6fR61

https://github.com/thirdiron/browzine-summon-adapter/pull/28/files#diff-2b01e2be49b9c65a2479fe3472439f6fR75

Notice what you describe is pretty much an end-to-end flow and touches multiple methods, perfect for an acceptance test.

Great question!

src/primo/browzine-primo-adapter.js
```diff
@@ -1,78 +1,290 @@
...
+  function searchResult($scope) {
+    var scope = getScope($scope);
+
+    if(!shouldEnhance(scope)) {
+      return;
+    }
+
+    var endpoint = getEndpoint(scope);
+
+    var request = new XMLHttpRequest();
```
tiandavis: For the time travelers to come, noting at this point in the timeline, jQuery was inaccessible in Primo's rollout of their Angular 1 web app for some reason. So it wasn't possible to do a `$.getJSON` or similar.

Issue #32, opened 2018-04-27
Need to opt-out of Browzine services


Steve-A-Orr:
As shared recently with Third Iron staff (Tian, Jordan, and John) in a technical support email chain for the Primo adapter, some libraries may desire to avoid displaying either the Article or Journal web links, as well as possibly cancel the use of the Browzine Cover Images.

Controlling the use of each service (two Web links and Cover Image) can be accomplished with the _non-mandatory_ introduction of only one variable (to the existing two variables) in the Primo **custom.js** file. 

I'll provide a pull request with the required adapter code changes. After loading the library:

- the proposed "primoUseCoverImageBrowZineService" variable needs to be set to **false** to avoid using cover images.
- either of the existing variables need to be set to an empty string ("") to stop showing web links for that type.

johnseguin:
Hi Steve, 

I'm closing this issue since in more recent versions we did add a similar functionality.  You can see where these flags are at (and we kept your idea going with other new features, always allowing them to be turned "off").  We recently added an option to turn on unpaywall data as well within the integration which you may bae interested in as well!

Thanks,
John

Steve-A-Orr:
I will certainly look at your improvements. Thank you for telling me.

PR #33, opened 2018-04-27
Browzine Service configurations (Issue #32)

Steve-A-Orr:
Summary - Github Issue #32 

Changes to adapter to control use of each Browzine service (Article web link, Journal web link, Cover images)

## Description

Changes made to two functions to check for whether a variable exists in Primo _custom.js_ file or whether legacy variables are an empty string, either of which brings new behavior into play.

## Implementation Notes

Entry of the following variables into the Primo _config.js_ file:

```
    primoJournalBrowZineWebLinkText: "View Journal",
    primoArticleBrowZineWebLinkText: "",
    primoUseCoverImageBrowZineService: false,
```
Using the values above, the expected result is that **no** _cover images_ are shown and only the journals will show a web link, in this case it would be the "View Journal" string.

## Deploy Precautions

- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -185,22 +185,51 @@ browzine.primo = (function() {
     var bookIcon = "https://assets.thirdiron.com/images/integrations/browzine_open_book_icon.png";
 
     if(isJournal(scope)) {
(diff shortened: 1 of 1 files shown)


karlbecker:
Hi Steve, thanks for the feedback and the PR.  Since there are other approaches to accomplishing this as mentioned in https://github.com/thirdiron/browzine-discovery-service-adapters/issues/32 , I'm going to close this without merging. Hope you're able to do the other things you wanted to now!

PR #36, opened 2018-05-21
remove link title

dermike:
The link title attribute here is unnecessary since it pretty much repeats the link text, but the main reason for this pull request is that the link title in the current form interferes with the ability to translate the link text.
We've translated the `primoJournalBrowZineWebLinkText` in our custom.js to Swedish, which is then ruined by the "in Browzine" in the title which we cannot adjust unless we host this script ourselves.

## Implementation Notes

1. ​

## Deploy Precautions

- :warning: Other: 
- None


karlbecker:
Thanks for this change @dermike , and sorry for the slow response time on this.  We have modified the script since your PR to remove the title attribute, so hopefully that takes care of your issue.

PR #37, opened 2018-06-19
feature/360-core-async-BZ-4772

tiandavis:
## Summary - [BZ-4772](https://thirdiron.atlassian.net/browse/BZ-4772)

Fixes core issue in 360 Core (EJP) where extra text like 'Terms of Use' appear in same block as result title and result in the target element not being found. Without the target element, we won't know which specific result to enhance. So results that didn't have "Terms of Use" next to the result title succeeded and those with "Terms of Use" next to the result title failed. This PR addresses and fixes this issue.

## Description

360 Core (EJP): 
Handles case where extra text like 'Terms of Use' appear in same block as result title.
Ensure Book Icon is property aligned.

Primo:
Removes Primo link title tag.

image

## Deploy Precautions

- None


tests/unit/models/360-core.js
```diff
@@ -1,6 +1,6 @@
 describe("SerSol 360 Core Model >", function() {
   var serSol360Core = {}, journalResponse = {}, scope = {}, titles = [], searchResults = {};
-  var results = "<div ui-view='searchResults'><div class='results-title-data'><div class='results-title-row'><div class='results-title-image-div'><img src='' ng-src='' class='results-title-image'/></div><div class='results-title-details'><div class='results-title'>The New England journal of medicine</div><div class='results-identifier'>ISSN: 0028-4793</div></div></div></div></div>";
+  var results = "<div ui-view='searchResults'><div class='results-title-data'><div class='results-title-row'><div class='results-title-image-div'><img src='' ng-src='' class='results-title-image'/></div><div class='results-title-details'><div class='results-title'>The Base Journal</div><div class='results-identifier'>ISSN: 0028-4793</div></div></div></div></div>";
```
karlbecker: _non-blocking PR comment_

When I saw this change, I thought, "oh, okay, base journal must be in the test somewhere," but now I see that this just gets replaced.  

Think it would make it a little clearer to rename to "The Base Journal Whose Name Gets Replaced" or something like that? Not sure, just thinking out loud.

tests/unit/models/360-core.js
```diff
@@ -232,6 +230,19 @@ describe("SerSol 360 Core Model >", function() {
...
+      var title = titles[0];
+      var journalName = $(title.target).find(".results-title").text();
+
+      $("div[ui-view='searchResults']").find(".results-title:contains('" + journalName + "')")[0].innerHTML += " Terms of Use";
+
+      var target = serSol360Core.getTarget(title);
+
+      expect(target).toBeDefined();
+      expect(target.innerHTML).toContain("The New England journal of medicine");
+      expect(target.innerHTML).toContain("The New England journal of medicine Terms of Use");
```
karlbecker: Good call!

Should we add a test for checking unexpected case changes, too?  I see the code does a `lowercase` on the title strings it compares, but I don't think we explicitly test for that, correct?

tests/unit/models/360-core.js
tiandavis: > "oh, okay, base journal must be in the test somewhere," but now I see that this just gets replaced.

The first thought was correct, no replacement. Updated because were were getting duplicate results because there is already a search result for "The New England journal of medicine" title in the test data: https://github.com/thirdiron/browzine-discovery-service-adapters/blob/develop/tests/unit/models/360-core.js#L19

tests/unit/models/360-core.js
tiandavis: Definitely. Added: https://github.com/thirdiron/browzine-discovery-service-adapters/pull/37/files#diff-bb9050811ef596c1707abe110a18bd52R249

Thanks!

PR #41, opened 2018-06-29
feature/circleci-2.0-migration-BZ-4835

tiandavis:
## Summary - [BZ-4835](https://thirdiron.atlassian.net/browse/BZ-4835)

CircleCI 2.0 Migration

## Description

image

## Deploy Precautions

- None


karlbecker:
For comparison purposes:
before-after-circleci2

Looks like we get potentially a tiny little time savings.

package.json
```diff
@@ -3,11 +3,11 @@
   "version": "1.0.0",
   "description": "Summon adapter to augment journal results with custom BrowZine API data.",
   "engines": {
-    "node": "0.10.32"
```
karlbecker: 😱 

👍

tiandavis:
Yeah, anywhere from 7 to 29 seconds. 😬 
The full build and deploy  cycle includes both the 00:46 + 01:25.

README.md
```diff
@@ -181,7 +181,7 @@ Enhancements and Bugfixes should be developed against the ES5 standard and teste
 ## Running / Development
 
 * `npm run server` (Runs a local webserver, visit http://localhost:8080)
-* `npm run tunnel` (Creates an ngrok tunnel, place this in the "Summon 2.0 External Script", e.g. https://9f9981c8.ngrok.io/src/summon/browzine-summon-adapter.js)
+* `npm install -g ngrok && ngrok http 8080` (Creates an ngrok tunnel, place this in the "Summon 2.0 External Script", e.g. https://9f9981c8.ngrok.io/src/summon/browzine-summon-adapter.js)
```
KeeganJ: Why did we make ngrok a peer dependency?

tiandavis: Really good question @KeeganJ! Ubuntu downloads an ngrok binary that slows down the build, so the motivation was to decrease build time. 

However, realizing we should probably keep the dependency local, move it and the `http-server` to `devDependencies`, then during the CircleCI run, we can do an `npm install --production` instead of only `npm install`. That should keep everything lean, using only what's necessary to build and deploy on CircleCI and maybe even speed up a little without the `http-server` install.

tiandavis:
Thanks Everyone!

README.md
KeeganJ: Sounds good to me!

PR #43, opened 2018-07-12
feature/primo-cycle-until-journal-cover-element-appears-BZ-4786

tiandavis:
## Summary - [BZ-4786](https://thirdiron.atlassian.net/browse/BZ-4786)

Primo: Polls for the Journal Cover Image element(s); Sets the Journal Cover Image element(s) once available.

## Description

Should be snappier overall. Should also work a lot better for internet connections that have lower bandwidth or higher latency issues.

## Deploy Precautions

- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -255,12 +255,14 @@ browzine.primo = (function() {
               var elementParent = getElementParent(element);
               var coverImages = elementParent.querySelectorAll("prm-search-result-thumbnail-container img");
 
-              if(coverImages[0].className.indexOf("fan-img") > -1) {
-                Array.prototype.forEach.call(coverImages, function(coverImage) {
-                  coverImage.src = coverImageUrl;
-                });
-              } else {
-                poll();
+              if(coverImages[0]) {
+                if(coverImages[0].className.indexOf("fan-img") > -1) {
+                  Array.prototype.forEach.call(coverImages, function(coverImage) {
+                    coverImage.src = coverImageUrl;
+                  });
+                } else {
+                  poll();
+                }
               }
             });
           })();
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -250,21 +250,19 @@ browzine.primo = (function() {
         }
 
         if(coverImageUrl && browzineEnabled) {
-          (function poll(){
-            setTimeout(function(){
-              var elementParent = getElementParent(element);
-              var coverImages = elementParent.querySelectorAll("prm-search-result-thumbnail-container img");
+          (function poll() {
+            var elementParent = getElementParent(element);
+            var coverImages = elementParent.querySelectorAll("prm-search-result-thumbnail-container img");
 
-              if(coverImages[0]) {
-                if(coverImages[0].className.indexOf("fan-img") > -1) {
-                  Array.prototype.forEach.call(coverImages, function(coverImage) {
-                    coverImage.src = coverImageUrl;
-                  });
-                } else {
-                  poll();
-                }
+            if(coverImages[0]) {
+              if(coverImages[0].className.indexOf("fan-img") > -1) {
+                Array.prototype.forEach.call(coverImages, function(coverImage) {
+                  coverImage.src = coverImageUrl;
+                });
+              } else {
+                requestAnimationFrame(poll);
               }
-            }, (1000 / 60));
+            }
           })();
         }
       }
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -264,7 +264,7 @@ browzine.primo = (function() {
                   poll();
                 }
               }
-            });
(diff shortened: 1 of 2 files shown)


src/primo/browzine-primo-adapter.js
```diff
@@ -247,21 +247,23 @@ browzine.primo = (function() {
...
         if(coverImageUrl && browzineEnabled) {
-          setTimeout(function() {
-            var elementParent = getElementParent(element);
-            var coverImages = elementParent.querySelectorAll("prm-search-result-thumbnail-container img");
-
-            Array.prototype.forEach.call(coverImages, function(coverImage) {
-              coverImage.src = coverImageUrl;
-              $scope.$apply();
+          (function poll(){
+            setTimeout(function(){
```
KeeganJ: If we aren't setting a timeout value here (a second parameter to the `setTimeout` function), I believe the value defaults to 0. Do we still need to wrap this function in a `setTimeout`? Or are we using it to emulate the behavior of `setImmediate`?

src/primo/browzine-primo-adapter.js
```diff
@@ -247,21 +247,23 @@ browzine.primo = (function() {
...
+              var elementParent = getElementParent(element);
+              var coverImages = elementParent.querySelectorAll("prm-search-result-thumbnail-container img");
+
+              if(coverImages[0].className.indexOf("fan-img") > -1) {
+                Array.prototype.forEach.call(coverImages, function(coverImage) {
+                  coverImage.src = coverImageUrl;
+                });
+              } else {
+                poll();
```
karlbecker: How frequently does this get called now?  Seems that we've got `setTimeout` being called without a time, and [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) it will do it "as soon as possible."

How much CPU does this take up?  I'm imagining this is using a decent amount of system resources...

src/primo/browzine-primo-adapter.js
tiandavis: > Do we still need to wrap this function in a setTimeout? 

Yes.

> Or are we using it to emulate the behavior of setImmediate?

Definitely the right ballpark here. Basically, the UI is updated asynchronously, so without the setTimeout, our code (which updates the UI), will run before the UI is ready. We don't actually need to wait any actual unit of time. What we really need to do is get our code in queue to execute after the display updates and any events ahead of us have completed.

For a very good article on microtask queueing and execution within the browser's event loop. See, https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

src/primo/browzine-primo-adapter.js
tiandavis: All good questions @karlbecker! The short answer is this will get called as many times as it needs to - No more, no less. The longer answer, of course, is the answer to this question is going to depend on a number of factors: Including the users network bandwidth, latency, their device's CPU, basically a host of things we can't control.

On my network and machine, I've seen Primo take anywhere from 1 to 3 (max) of these iterations to finally update their UI. Someone in Europe or New Zealand could have more iterations. Will they notice anything different, assuming their internet connection is generally just slow? Probably not. But whenever Primo updates their UI, we do the Journal Cover enhancements and the iterations stop.

We could add time to the intervals by setting a timeout value, of course, but the next customer support email we get about this will read something along the lines of the Journal Cover updates being "choppy", etc. :)

The key is what we have here is the fastest these Journal Cover enhancements can get. I'm comfortable trying out what we have here and then iterating on any feedback we get.

tiandavis: Fascinating stuff!

image

https://developer.mozilla.org/en-US/docs/Web/API/Frame_Timing_API
http://www.david-amador.com/2011/08/hurry-you-only-have-16-ms-to-render-everything/
https://www.quora.com/How-many-frames-per-second-can-a-browser-render-HTML

KeeganJ: screen shot 2018-07-12 at 5 50 04 pm

Ooooooo

screen shot 2018-07-12 at 5 50 31 pm

Awwwwww. That could have been really cool. But 16 or 15ms should still cover all 99% of our users 😄

tiandavis: Ok, we're pinned to 60fps now, but now ya got me wondering if we can use [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) here...?

karlbecker:
Hey @tiandavis ,

> and we don't take up any of those CPU cycles in between frames.

Yeah, this was the important piece for me - without this delay, we'd be making that CPU work potentially pretty hard. Thanks for the improvements!

What happens if an error happens? I'd like to make sure if someone opens a Primo page, and then something hiccups with their internet connection or our server, we don't eat significant CPU for the potentially hours or days they'll keep that tab open, forgetting about it but paying a cost on their CPU.

Maybe it's best just to test this to see what happens - we have a place to test it ready to go Tian?

Thanks!

PR #45, opened 2018-07-31
feature/direct-to-pdf-BZ-4508

tiandavis:
## Summary - [BZ-4508](https://thirdiron.atlassian.net/browse/BZ-4508)

Add Direct To PDF Link when available - Primo and Summon

## Description

Tasks:
- Add Direct To PDF Link when available - Primo (priority) && Summon (DONE)
Primo (DONE)
Summon (DONE)
Unit and Acceptance Tests (DONE)
Cross Browser Tests (Chrome, FF, Safari, IE11, Edge) (DONE)
- Summon: Refactor template to position "Article PDF" the first entry (DONE)
- Primo: Refactor template to align "Download PDF" with native "Available Online" Link (DONE)
- Override the default link text shown (DONE)
- On by default, but can disable with a config flag (DONE)
- Remove browzineEnabled check for CoverImages (DONE)
- Always update the Journal Cover when available, unless the Journal Cover is the default image (DONE)

**Screenshots**

Primo

image

Summon

image

## Deploy Precautions

- None

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ As a Custom Script Url:
 
 Customize the naming conventions for each type of search result - Journal/Article - by changing the wording in the quotes below:
 
-e.g. You can customize "View the Journal", "View Complete Issue", "Browse Now". These customizations are optional and the defaults are shown below.
+e.g. You can customize "View the Journal", "View Complete Issue", "Browse Now" and whether to enable direct to PDF links. These customizations are optional and the defaults are shown below.
 
 ```
 var browzine = {
@@ -67,6 +67,9 @@ var browzine = {
   summonArticleWording: "View Complete Issue",
   summonJournalBrowZineWebLinkText: "Browse Now",
   summonArticleBrowZineWebLinkText: "Browse Now",
+  summonArticlePDFDownloadLinkEnabled: true,
+  summonArticlePDFDownloadWording: "Article PDF",
+  summonArticlePDFDownloadLinkText: "Download Now",
 };
 ```
 
@@ -123,8 +126,6 @@ Update the following code snippet with the `api` endpoint and `apiKey` values:
   window.browzine = {
     api: "https://public-api.thirdiron.com/public/v1/libraries/XXX",
     apiKey: "ENTER API KEY",
-    primoJournalBrowZineWebLinkText: "View Journal Contents",
-    primoArticleBrowZineWebLinkText: "View Issue Contents",
   };
 
   browzine.script = document.createElement("script");
@@ -162,6 +163,8 @@ window.browzine = {
   apiKey: "ENTER API KEY",
   primoJournalBrowZineWebLinkText: "View Journal Contents",
   primoArticleBrowZineWebLinkText: "View Issue Contents",
+  primoArticlePDFDownloadLinkEnabled: true,
+  primoArticlePDFDownloadLinkText: "Download PDF",
 };
 ```
 
--- a/src/summon/browzine-summon-adapter.js
+++ b/src/summon/browzine-summon-adapter.js
@@ -183,14 +183,10 @@ browzine.summon = (function() {
     return enableShowDirectToPDFLink;
   };
 
-  function buildTemplate(scope, browzineWebLink, directToPDFUrl) {
+  function browzineWebLinkTemplate(scope, browzineWebLink) {
     var wording = "";
     var browzineWebLinkText = "";
     var bookIcon = "https://assets.thirdiron.com/images/integrations/browzine_open_book_icon.png";
-    var pdfIcon = "https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine_pdf_download_icon.png";
-    var articlePDFDownloadWording = "";
-    var articlePDFDownloadLinkText = "";
-    var directToPDFTemplate = "";
 
     if(isJournal(scope)) {
       wording = browzine.journalWording || browzine.summonJournalWording || "View the Journal";
@@ -200,17 +196,9 @@ browzine.summon = (function() {
     if(isArticle(scope)) {
       wording = browzine.articleWording || browzine.summonArticleWording || "View Complete Issue";
       browzineWebLinkText = browzine.articleBrowZineWebLinkText || browzine.summonArticleBrowZineWebLinkText || "Browse Now";
-
-      articlePDFDownloadWording = browzine.summonArticlePDFDownloadWording || "Article PDF";
-      articlePDFDownloadLinkText = browzine.summonArticlePDFDownloadLinkText || "Download Now";
-
-      if(showDirectToPDFLink() && directToPDFUrl) {
-        directToPDFTemplate = "{articlePDFDownloadWording}: <a class='browzine-direct-to-pdf-link' href='{directToPDFUrl}' target='_blank' style='text-decoration: underline; color: #333;'>{articlePDFDownloadLinkText}</a> <img class='browzine-pdf-icon' src='{pdfIcon}'/> <br/>";
-      }
     }
 
     var template = "<div class='browzine'>" +
-                     "{directToPDFTemplate}" +
                      "{wording}: <a class='browzine-web-link' href='{browzineWebLink}' target='_blank' style='text-decoration: underline; color: #333;'>{browzineWebLinkText}</a> <img class='browzine-book-icon' src='{bookIcon}'/>" +
                    "</div>";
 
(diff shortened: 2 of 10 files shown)


tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -169,11 +170,19 @@ describe("BrowZine Primo Adapter >", function() {
...
       expect(template).toBeDefined();
-      expect(template.text().trim()).toEqual("View Issue Contents");
+
+      expect(template.text().trim()).toContain("View Issue Contents");
+      expect(template.text().trim()).toContain("Download Now");
+
       expect(template.find("a.browzine-web-link").attr("href")).toEqual("https://browzine.com/libraries/XXX/journals/18126/issues/7764583?showArticleInContext=doi:10.1136/bmj.h2575");
       expect(template.find("a.browzine-web-link").attr("target")).toEqual("_blank");
-      expect(template.find("img.browzine-book-icon").attr("src")).toEqual("https://assets.thirdiron.com/images/integrations/browzine_open_book_icon.png");
+      expect(template.find("img.browzine-book-icon").attr("src")).toEqual("https://assets.thirdiron.com/images/integrations/browzine-open-book-icon.png");
```
karlbecker: dash-er-ized >> snake_case ? 😄

tests/unit/models/primo.js
```diff
@@ -617,7 +618,120 @@ describe("Primo Model >", function() {
...
+    });
+
+    it("should return true when a default coverImageUrl is returned by the API", function() {
+      var coverImageUrl = "https://assets.thirdiron.com/images/covers/default.png";
+      expect(primo.isDefaultCoverImage(coverImageUrl)).toEqual(true);
+    });
+  });
+
+  describe("primo model getDirectToPDFUrl method >", function() {
+    it("should not include a fullTextFile property in the BrowZine API response for a journal", function() {
```
karlbecker: I'd suggest a little rearrangement of the test here to match the given / when / then sort of format that is [popular with Cucumber](https://docs.cucumber.io/gherkin/reference/#steps) and generally [a BDD-style approach](https://martinfowler.com/bliki/GivenWhenThen.html).

How about something sort of like this?
```
describe("primo model getDirectToPDFUrl method >", function() {
  describe("given an API response for a journal", ...
    var scope = { ... issn: ["0096-6762" ...
    it("should not include a fullTextFile property", ...
```

I was at first unclear about where the test setup was and what the fixture was used for.  

After reading the test again, I can see the fixture is all about a journal response - and yup, journal responses don't have `fullTextFile` properties! 😊 

Typically we have fixtures setup in a `describe` setup, since the `describe` is usually a `given` or `when`.

What do you think of this approach, @tiandavis ?  Also, we could address this after this PR is merged if we'd like to get this deployed for John's demo tomorrow.

Thanks!

tests/unit/models/primo.js
```diff
@@ -617,7 +618,120 @@ describe("Primo Model >", function() {
...
+      };
+
+      var data = primo.getData(journalResponse);
+
+      expect(data).toBeDefined();
+
+      expect(primo.getDirectToPDFUrl(scope, data)).toBeNull();
+    });
+
+    it("should include a fullTextFile property in the BrowZine API response for an article", function() {
```
karlbecker: Same thing here - 
describe("primo model getDirectToPDFUrl method >", function() {
  describe("given an API response for a journal", ... 
  );
  describe("given an API response for an article", ...
    var scope = { ... doi: ["10.1136/bmj.h2575" ...
    it("should include a fullTextFile property", ...
```

Thoughts?

File: tests/unit/models/primo.js [hunk 4]
```diff
@@ -617,7 +618,120 @@ describe("Primo Model >", function() {
...
+    it("should show direct to pdf link when configuration property is undefined or null", function() {
+      expect(primo.showDirectToPDFLink()).toEqual(true);
+    });
+
+    it("should show direct to pdf link when configuration property is true", function() {
+      browzine.primoArticlePDFDownloadLinkEnabled = true;
+      expect(primo.showDirectToPDFLink()).toEqual(true);
+    });
+
+    it("should hide direct to pdf link when configuration property is false", function() {
```
karlbecker: 👍 to this unit test.

Do we want an acceptance-level test that checks to ensure there's no element that appears when this configuration property is false, too?  Similar to [this assertion](https://github.com/thirdiron/browzine-discovery-service-adapters/pull/45/files#diff-e918fc280a5c48d536bacb6a76a7de33R730), but we'd have a test and setup that makes sure the fully-output data truly excludes the PDF stuff.  

Basically, a test to ensure that `showDirectToPDFLink()` has an impact on the final HTML.

src/summon/browzine-summon-adapter.js
```diff
@@ -150,10 +150,60 @@ browzine.summon = (function() {
...
+    if(typeof config === "undefined" || config === null || config === true) {
+      enableShowDirectToPDFLink = true;
+    }
+
+    return enableShowDirectToPDFLink;
+  };
+
+  function directToPDFTemplate(directToPDFUrl) {
+    var pdfIcon = "https://s3.amazonaws.com/thirdiron-assets/images/integrations/browzine-pdf-download-icon.png";
+    var articlePDFDownloadWording = browzine.summonArticlePDFDownloadWording || "Article PDF";
```
karlbecker: _non-blocking PR comment_

Could be a potential improvement in the future, but what if we collected all these default strings and put them at the top of the file in one place?  Would let John or someone else very easily see, "oh, here are all the things that a user might see in the UI," and allow easier customization that way.

src/summon/browzine-summon-adapter.js
```diff
@@ -197,14 +246,20 @@ browzine.summon = (function() {
...
 
       if(browzineWebLink) {
-        var template = buildTemplate(scope, browzineWebLink);
+        var template = browzineWebLinkTemplate(scope, browzineWebLink);
         $(documentSummary).find(".docFooter .row:eq(0)").append(template);
       }
 
-      if(coverImageUrl && browzineEnabled) {
+      if(coverImageUrl && !defaultCoverImage) {
         $(documentSummary).find(".coverImage img").attr("src", coverImageUrl).attr("ng-src", coverImageUrl).css("box-shadow", "1px 1px 2px #ccc");
```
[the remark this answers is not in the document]
tiandavis: > is this an optimization to reduce the number of times we're trying to set the cover image

`browzineEnabled` stops the default cover image from appearing. But we want to expand Journal Cover coverage to show any Journal Covers we have available. The presence of the default cover image means we do not have the actual Journal Cover, hence the new `defaultCoverImage` check.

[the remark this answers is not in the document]
karlbecker: > But we want to expand Journal Cover coverage to show any Journal Covers we have available.

Right - in the past, we would show our default cover image in Primo/Summon, correct?  

But now, with this change, we will _not_ show our default cover image there, and we will just leave whatever Primo/Summon use by default... if I'm reading this right, anyway.

Am I grokking it? 🤔

tests/unit/models/primo.js
tiandavis: @karlbecker Fantastic question @karlbecker! TLDR; is this is how we'd want to think about Acceptance/Integration Tests, not necessarily Unit Tests. E.g. notice our [Acceptance Tests](https://github.com/thirdiron/browzine-discovery-service-adapters/blob/develop/tests/acceptance/browzine-primo-adapter-test.js#L6) have that high-level journal vs article organization. Whereas, our Unit Test here is more granularly specific to a methods and sometimes those methods test for both journal or article use. 

I could see updating the `it` statement to focus more on what the method should be doing and not so much on the API details though. Will update the same.

Good point on Cucumber though. Noting, I wouldn't knock a development team that used Cucumber for their Acceptance Tests since there is the aspect of bridging the gap between development and business requirements. But, I my mind, the biggest benefit is using it at the higher Integration Testing Environment level since these business requirements typically involve multiple applications integrating to succeed on that higher level business requirement: e.g. apps here, `browzine-discovery-service-adapters` and `browzine_cms`.

tests/unit/models/primo.js
tiandavis: Thanks Karl! See, https://github.com/thirdiron/browzine-discovery-service-adapters/pull/45#discussion_r208045501

README.md
```diff
@@ -67,6 +67,9 @@ var browzine = {
   summonArticleWording: "View Complete Issue",
   summonJournalBrowZineWebLinkText: "Browse Now",
   summonArticleBrowZineWebLinkText: "Browse Now",
+  summonArticlePDFDownloadLinkEnabled: true,
+  summonArticlePDFDownloadWording: "Article PDF",
```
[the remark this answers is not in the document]
tiandavis: Yeah, with these configuration properties, it's like, "sigh, people are already using them, so let's jus try our best and then support until these adapters get integrated into the host product officially."

src/summon/browzine-summon-adapter.js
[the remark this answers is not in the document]
tiandavis: > in the past, we would show our default cover image in Primo/Summon, correct? 

No. Because the `browzineEnabled` flag seemed to always come with an actual Journal Cover.

tests/unit/models/primo.js
tiandavis: Good idea Karl. Adding these now.

tiandavis:
Thanks for all the feedback folks!

PR #47, opened 2018-08-08
feature/circle-2-0-add-npm-cache-BZ-4890

tiandavis:
## Summary - [BZ-4890](https://thirdiron.atlassian.net/browse/BZ-4890)

Adds package caching.

## Deploy Precautions

- None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,13 +12,22 @@ build-environment: &build-environment
   name: Build Environment
   command: ~/project/.circleci/build.sh
 
+package-cache: &package-cache
+  key: package-cache-{{ checksum "package.json" }}
+
 jobs:
   build:
     <<: *settings
     steps:
       - checkout
+      - restore_cache:
+          <<: *package-cache
       - run:
           <<: *build-environment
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -70,8 +91,14 @@ jobs:
     <<: *settings
     steps:
       - checkout
+      - restore_cache:
```
karlbecker: Do you need the restore and save caches of every one of these deploys if we already do it in the `build` job?  

I don't quite know how the caches work there, but seems like the restore and save cache we do in the build phase might stick around until these phases... but perhaps not, since these phases will be performed in separate Docker containers, won't they?

tiandavis: Every job run is isolated. So assets from the build job don't remain when the deploy job is ran. Since we're using an npm/node package to deploy, we need to either `npm install` or restore the cache.

Now you see why I tried to not sure it initially, but it's for the better now. I'm TOTALLY not a fan of how CircleCI did this. It definitely broke the Principle of Least Surprise for me too.

PR #51, opened 2018-08-23
direct-to-pdf-configs-prefix-support-BZ-4508

tiandavis:
## Summary - [BZ-4508](https://thirdiron.atlassian.net/browse/BZ-4508)

Adds support for summon/primo prefixed configs. 

## Deploy Precautions

- None

### DIFF
--- a/src/summon/browzine-summon-adapter.js
+++ b/src/summon/browzine-summon-adapter.js
@@ -174,7 +174,7 @@ browzine.summon = (function() {
 
   function showDirectToPDFLink() {
     var enableShowDirectToPDFLink = false;
-    var config = browzine.articlePDFDownloadLinkEnabled;
+    var config = browzine.summonArticlePDFDownloadLinkEnabled || browzine.articlePDFDownloadLinkEnabled;
 
     if(typeof config === "undefined" || config === null || config === true) {
       enableShowDirectToPDFLink = true;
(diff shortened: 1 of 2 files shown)


karlbecker:
Just to clarify - are we adding these back in because they already exist out in the wild?

johnseguin:
@karlbecker  Yes.  This is @tiandavis helping me save face on the comms side. :)

src/summon/browzine-summon-adapter.js
```diff
@@ -174,7 +174,7 @@ browzine.summon = (function() {
 
   function showDirectToPDFLink() {
     var enableShowDirectToPDFLink = false;
-    var config = browzine.articlePDFDownloadLinkEnabled;
+    var config = browzine.summonArticlePDFDownloadLinkEnabled || browzine.articlePDFDownloadLinkEnabled;
```
karlbecker: How about we have the platform-specific variable consistently either on the left or right side of the `||` ?  That way, we can clearly state which item takes precedence if anyone has some funny configuration.

src/summon/browzine-summon-adapter.js
```diff
@@ -185,8 +185,8 @@ browzine.summon = (function() {
 
   function directToPDFTemplate(directToPDFUrl) {
     var pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-pdf-download-icon.svg";
-    var articlePDFDownloadWording = browzine.articlePDFDownloadWording || "Article PDF";
-    var articlePDFDownloadLinkText = browzine.articlePDFDownloadLinkText || "Download Now";
+    var articlePDFDownloadWording = browzine.articlePDFDownloadWording || browzine.summonArticlePDFDownloadWording || "Article PDF";
+    var articlePDFDownloadLinkText = browzine.articlePDFDownloadLinkText || browzine.summonArticlePDFDownloadLinkText || "Download Now";
```
karlbecker: Is there a test for the `summonArticlePDFDownloadWording` ?

tiandavis: Nope. Will add.

src/summon/browzine-summon-adapter.js
tiandavis: Awww, Hey @karlbecker! Just remembered why I didn't do that. The nonprefix was the original, so we have those out there. Then prefix was the second high-level iteration. Returning to the non-prefix should probably take the precedence going forward.

It's just this particular check is a weird case. Let's see if the new tests reveal anything.

tiandavis:
Good catches @karlbecker! Updates look good.

PR #53, opened 2018-10-01
feature/primo-user-customizations-BZ-4967

tiandavis:
## Summary - [BZ-4967](https://thirdiron.atlassian.net/browse/BZ-4967)

Adds new customization properties, including and option to not enhance print records.

## Implementation Notes

1. ​

## Deploy Precautions

- :warning: Other: 
- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -106,10 +106,6 @@ browzine.primo = (function() {
     return endpoint;
   };
 
-  function shouldEnhance(scope) {
-    return !!((isJournal(scope) && getIssn(scope)) || (isArticle(scope) && getDoi(scope)));
-  };
-
   function getData(response) {
     var data = {};
 
@@ -202,20 +198,81 @@ browzine.primo = (function() {
     return directToPDFUrl;
   };
 
+  function showJournalCoverImages() {
+    var featureEnabled = false;
+    var config = browzine.journalCoverImagesEnabled;
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
+    }
+
+    return featureEnabled;
+  };
+
+  function showJournalBrowZineWebLinkText() {
+    var featureEnabled = false;
+    var config = browzine.journalBrowZineWebLinkTextEnabled;
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
+    }
+
+    return featureEnabled;
+  };
+
+  function showArticleBrowZineWebLinkText() {
+    var featureEnabled = false;
+    var config = browzine.articleBrowZineWebLinkTextEnabled;
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
+    }
+
+    return featureEnabled;
+  };
+
   function showDirectToPDFLink() {
-    var enableShowDirectToPDFLink = false;
+    var featureEnabled = false;
     var config = browzine.articlePDFDownloadLinkEnabled;
     var prefixConfig = browzine.primoArticlePDFDownloadLinkEnabled;
 
     if(typeof config === "undefined" || config === null || config === true) {
-      enableShowDirectToPDFLink = true;
+      featureEnabled = true;
     }
 
     if(typeof prefixConfig !== "undefined" && prefixConfig !== null && prefixConfig === false) {
-      enableShowDirectToPDFLink = false;
+      featureEnabled = false;
     }
 
-    return enableShowDirectToPDFLink;
+    return featureEnabled;
+  };
+
+  function showPrintRecords() {
+    var featureEnabled = false;
+    var config = browzine.printRecordsIntegrationEnabled;
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
+    }
+
+    return featureEnabled;
+  };
+
+  function isFiltered(scope) {
+    var validation = false;
+    var result = getResult(scope);
+
+    if(result && result.delivery) {
+      if(result.delivery.deliveryCategory) {
+        var deliveryCategory = result.delivery.deliveryCategory[0].trim().toLowerCase();
+
+        if(deliveryCategory === "alma-e" && !showPrintRecords()) {
+          validation = true;
+        }
+      }
+    }
+
+    return validation;
   };
 
   function directToPDFTemplate(directToPDFUrl) {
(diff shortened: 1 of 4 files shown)


src/primo/browzine-primo-adapter.js
```diff
@@ -276,6 +333,10 @@ browzine.primo = (function() {
     return $scope && $scope.$ctrl && $scope.$ctrl.parentCtrl ? $scope.$ctrl.parentCtrl : undefined;
   };
 
+  function shouldEnhance(scope) {
+    return !!(!isFiltered(scope) && ((isJournal(scope) && getIssn(scope)) || (isArticle(scope) && getDoi(scope))));
```
KeeganJ: Could we break out the logic on this line into separate steps? I had some difficulty parsing it with the level of parentheses nesting. A couple alternatives that helped me understand the logic:

Split the line:

```js
return !!(
  !isFiltered(scope)
  && (
    (
      isJournal(scope)
      && getIssn(scope)
    )
    || (
      isArticle(scope)
      && getDoi(scope)
    )
  )
);
```

Or use short circuit logic (this should be functionally equivalent):

```js
if (isFiltered(scope)) return false;

if (isJournal(scope) && getIssn(scope)) return true;

if (isArticle(scope) && getDoi(scope)) return true;

return false;
```

src/summon/browzine-summon-adapter.js
```diff
@@ -236,6 +284,10 @@ browzine.summon = (function() {
     return angular.element(documentSummary).scope();
   };
 
+  function shouldEnhance(scope) {
+    return !!(!isFiltered(scope) && ((isJournal(scope) && getIssn(scope)) || (isArticle(scope) && getDoi(scope))));
```
KeeganJ: Ditto the above

src/primo/browzine-primo-adapter.js
tiandavis: Yep! It's time. Will look into it. Thanks!

karlbecker: Agreed, a little reformatting here could speed up comprehending this logic.

src/primo/browzine-primo-adapter.js
```diff
@@ -202,20 +198,81 @@ browzine.primo = (function() {
...
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
     }
 
-    return enableShowDirectToPDFLink;
+    return featureEnabled;
+  };
+
+  function isFiltered(scope) {
```
karlbecker: By `isFiltered`, what are we filtering on? Looks like this is specifically for print records, but is it more than that?

Maybe a little tweak to the function name could make it a little quicker to figure that out. Perhaps `isFilteredOutDueToPrint` if it's just based on print?

tests/unit/models/360-core.js
```diff
@@ -165,6 +165,54 @@ describe("SerSol 360 Core Model >", function() {
...
+    });
+
+    it("should show journal cover when configuration property is true", function() {
+      browzine.journalCoverImagesEnabled = true;
+      expect(serSol360Core.showJournalCoverImages()).toEqual(true);
+    });
+
+    it("should not show journal cover when configuration property is false", function() {
+      browzine.journalCoverImagesEnabled = false;
+      expect(serSol360Core.showJournalCoverImages()).toEqual(false);
```
karlbecker: Glad we've got some tests here! However, should we add a test that ensures when `serSol360Core.showJournalCoverImages()` is false, the cover image truly does not show up in the record?  Perhaps a test similar to [this one](https://github.com/thirdiron/browzine-discovery-service-adapters/blob/dd8eeb223016c8aeebf952afba98d8ee7154d109/tests/unit/models/360-core.js#L407), except checking for the image to not appear?

tests/unit/models/primo.js
```diff
@@ -713,6 +785,112 @@ describe("Primo Model >", function() {
...
+          delivery: {
+            deliveryCategory: ["Alma-P"]
+          }
+        }
+      };
+      browzine.printRecordsIntegrationEnabled = false;
+      expect(primo.isFiltered(scope)).toEqual(true);
+    });
+  });
+
```
karlbecker: Same here - maybe a test for primo to not display journal cover images and BZ Web links in the actual DOM element we construct when those settings are set to false?

tests/unit/models/360-core.js
```diff
@@ -165,6 +165,54 @@ describe("SerSol 360 Core Model >", function() {
...
+    });
+  });
+
+  describe("serSol360Core model showJournalBrowZineWebLinkText method >", function() {
```
karlbecker: Same with this - we check that `serSol360Core.showJournalBrowZineWebLinkText()` is true or false properly, but we don't test that `serSol360Core.showJournalBrowZineWebLinkText()` influences what appears in the DOM from what I can see.  If we add that test, we'll be really tight on our coverage though 👍

tests/unit/models/summon.js
```diff
@@ -362,6 +362,78 @@ describe("Summon Model >", function() {
     });
   });
 
+  describe("summon model showJournalCoverImages method >", function() {
```
karlbecker: Same for adding a couple tests for Summon's DOM elements behaving per the flag like I mentioned for the other two.

tiandavis:
Really good feedback folks! On it. Thanks!

tests/unit/models/primo.js
tiandavis: These Unit tests can only test layout we produce: e.g. `directToPDFTemplate` and `browzineWebLinkTemplate`. We don't have one for Journal Covers because we're able to inject into the exiting Primo template: https://github.com/thirdiron/browzine-discovery-service-adapters/blob/develop/src/primo/browzine-primo-adapter.js#L330

But definitely get it though, these "DOM" tests would be at the Acceptance Test level since they require the whole UI down to run. So instead of specific functions, we'd need to setup a top-level scenario for each of those flag tests. Taking a look...

tiandavis: Yep. Checkout this Acceptance Test:
https://github.com/thirdiron/browzine-discovery-service-adapters/blob/develop/tests/acceptance/browzine-primo-adapter-test.js#L203

Something like that. Working through these new ones now. Thanks!

karlbecker: Thanks @tiandavis ! Hopefully it's not _too_ painful, but it will probably help us avoid any nasty regressions in the future 😊

tiandavis: [Acceptance Tests](https://github.com/thirdiron/browzine-discovery-service-adapters/pull/53/commits/5eb15828dc0a3efcfca5c2b0caae630a18c3df68) or the customization properties are added. Should be all set now @karlbecker!

PR #59, opened 2019-08-13
feature/adds-icon-alt-tags-5453

tiandavis:
## Summary - [BZ-5453](https://thirdiron.atlassian.net/browse/BZ-5453)

Adds alt tag to img elements.

## Deploy Precautions

- None


tiandavis:
> By the way, now that you've used React for the Nomad buttons, how much do you want to use it in this now? 😊

image

I can't wait to use React here all of **zero** times. 😂

> (although I have no idea if it's easy, or if it would even make sense)

The difficulty may be more in the product strategy than the tech. LibKey Nomad is a packaged solution, so there probably isn't an expectation of readable code. Whereas, the discovery service adapters are closer to SDKs where another developer or technical resource has to implement it into their existing adapter addons. Because the adapter implementation is more hands on, library developers may want to peak behind the covers and see what they're injecting into their discovery service. If we added a view library like React, we'd likely absolutely need to add a bundler like Webpack to manage all the dependencies that come with a React installation. Once we do that, should anyone peak behind the covers, they'd go from seeing a manicured garden to really an endless endless mirage of code.

Not saying we should never add React here though...

Probably the best indication of the right time to add React would be if the business asks us to add more complexity to the adapters. e.g. flyout panels, popups/modals, custom/more interactive UI/UX, any kind of complex dropdowns like we have in the Nomad popup; Think "complex" UI or interactions generally. Then we'd have to reset expectations with the business and let them know that we probably can't provide readable unminified code any more because of the dependencies we'd have to add to accomplish a more interactive and complex UI/UX.

Makes sense?

PR #67, opened 2020-01-29
feature/intercept-link-events-BZ-5771

tiandavis:
## Summary - [BZ-5771](https://thirdiron.atlassian.net/browse/BZ-5771)

Intercepts Browzine link click events to ensure response consistency.

## Description

We’ve seen some discovery services intercept basic a href links, and have been encouraged to intercept clicks more closely. We should continue intercepting clicks like this unless we hear feedback from discovery service vendors that this is no longer desired or necessary.

## Deploy Precautions

- None

### DIFF
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -183,6 +183,15 @@ browzine.serSol360Core = (function() {
     return !!issn;
   };
 
+  function transition(event, anchor) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    window.open(anchor.href, anchor.target);
+
+    return false;
+  };
+
   function browzineWebLinkTemplate(browzineWebLink) {
     var browzineWebLinkText = "";
     var bookIcon = "https://assets.thirdiron.com/images/integrations/browzine-open-book-icon.svg";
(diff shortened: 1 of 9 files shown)


src/360-core/browzine-360-core-adapter.js
```diff
@@ -191,7 +200,7 @@ browzine.serSol360Core = (function() {
 
     var template = "<div class='browzine' style='margin: 5px 0;'>" +
                      "<img alt='BrowZine Book Icon' class='browzine-book-icon' src='{bookIcon}' style='margin-top: -3px; display: inline;' width='16' height='15'/> " +
-                     "<a class='browzine-web-link' href='{browzineWebLink}' target='_blank' style='font-weight: 300;'>{browzineWebLinkText}</a>" +
+                     "<a class='browzine-web-link' href='{browzineWebLink}' target='_blank' style='font-weight: 300;' onclick='browzine.serSol360Core.transition(event, this)'>{browzineWebLinkText}</a>" +
```
karlbecker: 👍 to adding this behavior to all 3 services we support for maximum future-proofing! 

It'll involve a little extra testing now, but would save hours in the future if a change ever makes this necessary.

tests/unit/models/360-core.js
```diff
@@ -411,6 +411,22 @@ describe("SerSol 360 Core Model >", function() {
...
+
+      serSol360Core.transition({
+        preventDefault: function() {},
+        stopPropagation: function() {}
+      }, {
+        href: "https://browzine.com/libraries/XXX/journals/10292",
+        target: "_blank"
+      });
+
+      expect(window.open).toHaveBeenCalledWith("https://browzine.com/libraries/XXX/journals/10292", "_blank");
```
karlbecker: Nice test ensuring that window.open is called! 👍

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -396,6 +396,18 @@ describe("BrowZine Primo Adapter >", function() {
           done();
         });
       });
+
+      it("should open a new window when a direct to pdf link is clicked", function() {
```
karlbecker: I'm wondering if a comment could be helpful to add here. I could easily see myself or someone else coming in here in the future and going, "Why are we using window.open? A simple `a` tag with the `target` set appropriately should be fine, let me refactor this and remove this test."

Either giving a reason for why the test is here, or giving a reason in the code for why we're doing that, would probably be helpful to avoid an accidental regression in the future.

tiandavis:
> do you think it's wise to put a copy of the HTML that necessitated this change into the repo, along with a test to confirm we interact with it properly?

It would have to be Primo Angular code that necessitated the change. AFAIK, not something we'd be able to add here.

PR #76, opened 2020-04-28
feature/unify-journal-covers-fix-BZ-5916

tiandavis:
## Summary - [BZ-5916](https://thirdiron.atlassian.net/browse/BZ-5916)

Unifies the Primo journal covers fix to work in both the sandbox and production environments.

## Description

Validated this in both production and sandbox.

## Deploy Precautions

- None


karlbecker:
I used works twice in that last sentence - clearly I'm sleepy 😄 

🛌

PR #77, opened 2020-05-07
Remove "printRecordsIntegrationEnabled" config line from Summon

johnseguin:
## Description

This is a Primo only feature.  While this won't *hurt* anything it will likely cause confusion for Summon admins wondering what it does.  :)  Removed to avoid confusion.

## Deploy Precautions

- None

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,6 @@ var browzine = {
   articleLinkTextWording: "Article Link",
   articleLinkText: "Read Article",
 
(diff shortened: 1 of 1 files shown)


karlbecker:
I think when we get changes like this @tiandavis , once we approve it we can merge them in for @johnseguin - I don't think he's much of a GitHub merger 😊 

Merging now!

PR #79, opened 2020-05-27
feature/summon-remove-link-colons-BZ-6021

tiandavis:
## Summary - [BZ-6021](https://thirdiron.atlassian.net/browse/BZ-6021)

Update the Summon integration to remove the colons in the text

## Description

Sample:
http://msulibrariestest.summon.serialssolutions.com/#!/search?ho=t&fvf=ContentType,Journal%20Article,f&l=en&q=roman%20empire

image

## Deploy Precautions

- None


tiandavis:
Looks like Summon officially removed the colon:
image

Pretty sure we're good to just remove it altogether. 
I just get nervous if we have to add it back because those were a lot of updates! 
But no point worrying about that now! 😂

PR #82, opened 2020-07-01
feature/unpaywall-config-updates-BZ-6097

tiandavis:
## Summary - [BZ-6097](https://thirdiron.atlassian.net/browse/BZ-6097)

Adds Unpaywall config gates.

## Description

TODO:
- Implementation (DONE)
- Acceptance Tests (DONE)

## Deploy Precautions

- None

### DIFF
--- a/tests/acceptance/browzine-primo-adapter-test.js
+++ b/tests/acceptance/browzine-primo-adapter-test.js
@@ -1166,6 +1166,42 @@ describe("BrowZine Primo Adapter >", function() {
           expect(template.find("a.unpaywall-article-pdf-link").attr("target")).toEqual("_blank");
           expect(template.find("img.browzine-pdf-icon").attr("src")).toEqual("https://assets.thirdiron.com/images/integrations/browzine-pdf-download-icon.svg");
         });
+
+        it("should not show an unpaywall article pdf when articlePDFDownloadViaUnpaywallEnabled is false", function() {
+          browzine.articlePDFDownloadViaUnpaywallEnabled = false;
+
+          var request = jasmine.Ajax.requests.mostRecent();
+
+          request.respondWith({
+            status: 200,
+            contentType: "application/json",
+            response: JSON.stringify({
+              "best_oa_location": {
+                "endpoint_id": null,
+                "evidence": "open (via free pdf)",
+                "host_type": "publisher",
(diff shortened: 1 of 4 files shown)


tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -1486,6 +1630,42 @@ describe("BrowZine Primo Adapter >", function() {
...
+        it("should not show an unpaywall article pdf when articlePDFDownloadViaUnpaywallEnabled is false", function() {
+          browzine.articlePDFDownloadViaUnpaywallEnabled = false;
+
+          var request = jasmine.Ajax.requests.mostRecent();
+
+          request.respondWith({
+            status: 200,
+            contentType: "application/json",
+            response: JSON.stringify({
+              "best_oa_location": {
```
karlbecker, line 1643: I don't think I've mentioned it before, but nice job on these test formats 👏 
Seeing the Unpaywall response in line with the test makes it easy to confirm if the test fixture is setup right.

I setup some tests in Nomad awhile ago, and I put the API responses in separate files. I was thinking the isolation would be helpful, but after reading these tests, I think this might be the way to go in Nomad, too. 

One difference between the two tests: Nomad has the full API response from Unpaywall, but here we have an abbreviated response. But I think the abbreviated response is perfectly fine for testing, too.

Just musing out loud here - curious what you think Tian!

tiandavis: > Just musing out loud here - curious what you think Tian!

FWIW, I think the way you setup the Nomad tests works just fine.

Testing infrastructure plays a part here as well. In Nomad, we have an express server emulating the 3i and Unpaywall API responses. Here, I'm leveraging Jasmine's ability to intercept and mock HTTP requests. So to get a similar inline effect in Nomad, we may have to find a corollary to Jasmine's ajax mocking, but in Mocha. I'm just not sure what that approach would look like or how well it would work out.

karlbecker: Good point @tiandavis , thanks for reminding me - Nomad requires a little heavier-duty way to serve up that fixturized data!

So yeah, no change for now in Nomad sounds like a plan to me 👍

PR #85, opened 2020-12-18
feature/article-link-by-default-BZ-6411

tiandavis:
## Summary - [BZ-6411](https://thirdiron.atlassian.net/browse/BZ-6411)

Show article link by default.

## Deploy Precautions

- None

### DIFF
--- a/src/summon/browzine-summon-adapter.js
+++ b/src/summon/browzine-summon-adapter.js
@@ -383,7 +383,7 @@ browzine.summon = (function() {
     var featureEnabled = false;
     var config = browzine.articleLinkEnabled;
 
-    if (config === true) {
+    if (typeof config === "undefined" || config === null || config === true) {
       featureEnabled = true;
     }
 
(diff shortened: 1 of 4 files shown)


src/summon/browzine-summon-adapter.js
```diff
@@ -383,7 +383,7 @@ browzine.summon = (function() {
     var featureEnabled = false;
     var config = browzine.articleLinkEnabled;
 
-    if (config === true) {
+    if (typeof config === "undefined" || config === null || config === true) {
```
karlbecker, line 386: Works for me 👍 

We could also just change it to check if `config === false` perhaps? But no biggie, this definitely gets the job done, and should be sure not to break the existing installs.

karlbecker (approved):
Simple change, LGTM @tiandavis !

src/summon/browzine-summon-adapter.js
tiandavis: > We could also just change it to check if config === false perhaps?

One would think, but alas JavaScript...
image

The current setup helps us account for when the `articleLinkEnabled` property doesn't exist at all. e.g `undefined` mainly, but possibly even a `null` value.

karlbecker: JavaScript!
iu-16

Thanks for ensuring we didn't ship my wrong idea 😄

PR #87, opened 2021-04-01
Use Node 12 for building and in package.json

karlbecker:
## Summary - [BZ-6592](https://thirdiron.atlassian.net/browse/BZ-6592)

Make sure this repo is no longer relying on an older version of node

## Description

Node 10 LTS is about to lose support, so let's at least move up to Node 12 LTS.  I would have moved to Node 14 LTS but we don't have a branch of our Docker image on 14 yet.

## Deploy Precautions

- None


tiandavis (approved):
@karlbecker Tests pass and code looks good! 🛳 🎉

PR #97, opened 2021-12-03
Summon Dec 2021 button and label GUI update - bz 6936

karlbecker:
## Summary - [BZ-6936](https://thirdiron.atlassian.net/browse/BZ-6936)

Uses the Summon CSS classes to add blue bubbles around links and a few other little adjustments to better match the upcoming Summon UI changes

## Description

Dec2021Update is turned off by default, but can be enabled in the config - see the README

Before on the left, after on the right (activated by putting `dec2021Update: true` into the config file):

before-after-summon2021ui

## Implementation Notes

1. I embedded the svg data right into the script for the PDF icon.  Think I should change that, and instead just make a file on S3?  
    * Alternatively, I could embed the svg data of all of our icons into this file, too
2. I had to add a space to [make things align up](https://thirdiron.slack.com/archives/C6ND59F7C/p1638905708033300) in the same way they do in the new Summon 2021 UI, so I had to make some changes in the last commit - https://github.com/thirdiron/browzine-discovery-service-adapters/pull/97/commits/6753595c9388ecc82aa8fcc0ab3ecb37c981d96b - to factor that in:
    *  image
    * Think I should use regexes with variable whitespace in them rather than two spaces like in there?  Or just modify the implementation to add a little extra padding in CSS to get the buttons to line, even though the space is what really is present in Summon 2021's UI?

## Deploy Precautions

- ⚠️ When this is merged, no change to the Summon UI will happen - we'll need to flip the `dec2021Update` toggle to `true`


karlbecker:
Hey @tiandavis ,
This is ready to review, although I made a noticeable change since I first marked it ready for review: I added unit test specifically for Summon Dec 2021 behavior.

Since I have Summon Dec 2021-specific tests and code in here, I also created [BZ-6951](https://thirdiron.atlassian.net/browse/BZ-6951) to remove the old Summon code once we the new Summon is fully rolled out and we no longer need to support the old Summon.

Let me know if you have any questions - thanks Tian!

tiandavis (changes requested):
Hi @karlbecker! The big high-level that stands out is the use of `browzine.dec2021Update` and the corresponding separate unit test and acceptance test files.

This is the way forward and release will be shortly next week. Shouldn't we just treat this as the forth coming baseline and just stabilize the conversion here. Thinking we don't even really need to merge this to `develop`. Just wait until we're green light to get this to IT then merge. In the mean time, related JIRAs can be branched off this PR.

Just thinking we need to get this baseline ready without the `browzine.dec2021Update` type code because it's not like a customer is going to set this `dec2021Update` property. Let me know if you want to hop on a call to weigh this. Or if this makes sense, I can help out and remove the `browzine.dec2021Update` references. Of course, you can handle that too. Let me know. Thanks!

karlbecker:
Hey @tiandavis ,
I kept the ability to go back to the old version as a "just in case" measure, since we can't control what Summon might do.  

Although December 15 is when things are supposed to ship out, since it's a third party that we have no control over, I wanted to keep an eject button in place in case we want to go back to the old way of doing things.  (say they roll out the new UI, a bunch of people complain, and then they undo it for a month or two while they iron out issues)

The change I planned to make (sorry, I didn't document this in the PR - I should have put that somewhere!) before Dec 14, then we could ship on Dec 14, is to just add the line `browzine.dec2021Update = true;` at the top of the file:
CleanShot 2021-12-09 at 09 34 14@2x

That means no updates needed for users - we just push a small change ourselves, and force it for everyone.

I created [BZ-6951](https://thirdiron.atlassian.net/browse/BZ-6951) to get rid of the old tests - I was thinking just basically renaming `browzine-summon-adapter-dec2021Update-test.js` to `browzine-summon-adapter-test.js` , and that it would be it, once we're confident Summon has rolled out their new UI and won't go backwards.

Then we can very easily toggle from having the new UI to not, since I don't 100% know if Summon will do as they've said they will.  Basically keep an easy exit in case Summon UI updates don't go as planned, to avoid us getting backed into any sort of unforeseen weird corner 😊 

Then we can get rid of the `oldSummon` variables and `dec2021Update` once we're confident Summon is all done and moved to the new UI.

What do you think @tiandavis ?

tiandavis (approved):

tiandavis:
Sounds like a plan @karlbecker!

PR #99, opened 2021-12-09
Add eslint

karlbecker:
## Summary - [BZ-XXXX](https://thirdiron.atlassian.net/browse/BZ-XXXX)

Adds eslint to this repo to potentially reduce the chance of undesired behavior sneaking in

## Description

This is just a draft for now, but wanted to get your opinion on adding eslint to the repo.  Do you like this setup so far @tiandavis ?

I didn't address all the eslint warnings and errors.  Instead I just took care of the majority of them by adding a few globals and specifying a couple of values in `env`

No need to look at this while you're finishing up the Summon 2021 changes - this is all for merging after we get those changes out. But since I was just in here I wanted to give a quick crack at this. 

Let me know what you think sometime after Summon 2021 is ready for review - thanks!

Also: think we want to add a package-lock.json file to this repo?

## Deploy Precautions

- None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,11 @@ jobs:
           paths:
             - node_modules
       - run:
+          name: Lint
+          command: |
+            . ~/.nvm/nvm.sh
+            npm run lint
+      - run:
           name: Automated Tests
           command: |
             . ~/.nvm/nvm.sh
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,52 @@
+{
+  "globals": {
+    "DEV_MODE": true,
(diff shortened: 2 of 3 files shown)


tiandavis:
Adding the `package-lock.json` sounds good @karlbecker! I've learned my lesson with React breaking because a downstream package broke ALL THE THINGS. 👍

IIRC, there was some issue with `const` and `let` because of the operating environment. I wonder if there is a rule we can create for eslint to stop the build if those are added. Until we can use a build system with babel, we still have to validate the es6 functions we choose have good cross browser compatibility, but not sure if this is something eslint can help us with.

Thank for getting this going! 🎉

PR #109, opened 2022-01-10
feature/libkey-link-optimizer-updates-BZ-6986

tiandavis:
## Summary - [BZ-6986](https://thirdiron.atlassian.net/browse/BZ-6986)

* Optimize links when either PDF or article link available
* Ensure only Third Iron links are on the line labeled "View Now", and all Summon-created links are on the "Journal Article" line
* Only show one link by default - as [mentioned in Slack](https://thirdiron.slack.com/archives/C6ND59F7C/p1642087165023800) and captured in [BZ-6943](https://thirdiron.atlassian.net/browse/BZ-6943)

## Deploy Precautions

- None

### DIFF
--- a/src/summon/browzine-summon-adapter.js
+++ b/src/summon/browzine-summon-adapter.js
@@ -830,13 +830,13 @@ browzine.summon = (function() {
           $(documentSummary).find(".coverImage img").attr("src", coverImageUrl).attr("ng-src", coverImageUrl).css("box-shadow", "1px 1px 2px #ccc");
         }
 
-        // if (!showLinkResolverLink() && (directToPDFUrl || articleLinkUrl)) {
-        //   var contentLinkElement = $(documentSummary).find(".availabilityContent");
-        //
-        //   if (contentLinkElement) {
-        //     contentLinkElement.remove();
-        //   }
-        // }
+        if (!showLinkResolverLink() && (directToPDFUrl || articleLinkUrl)) {
+          var contentLinkElement = $(documentSummary).find(".availabilityContent");
+
+          if (contentLinkElement) {
+            contentLinkElement.remove();
+          }
+        }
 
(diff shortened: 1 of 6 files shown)


karlbecker (approved):
LGTM @tiandavis ! I haven't run it yet to test it out, but the changes I'm seeing make it look like it should work smoothly.

Note that I added a little detail to the PR description, hopefully that captures the three main changes we've got going here.

src/summon/browzine-summon-adapter.js
```diff
@@ -856,12 +838,21 @@ browzine.summon = (function() {
...
-          }
+          (function poll() {
+            var quickLinkElement = $(documentSummary).find(".docFooter .availabilityFullText a[display-text='::i18n.translations.PDF']");
+
+            if (quickLinkElement.length > 0) {
+              quickLinkElement.remove();
+            } else {
+              if (intervals < 120) {
+                intervals++;
+                requestAnimationFrame(poll);
```
karlbecker, line 852: One thought: instead of requestAnimationFrame, what if we did this with a setTimeout and only checked, say, at 100 msec increments for up to 20 seconds or something like that?

When I looked at other uses of requestAnimationFrame (like in Nomad), it seems like it goes really fast (I think firing around once per msec), although especially here I think we only need to check at a relatively infrequent rate (like once per 100 msec).

Thoughts?

karlbecker (approved):
One more thought @tiandavis - we can test this out in staging and see how it goes, but I wonder if we would want to use setTimeout or something that fires less frequently than requestAnimationFrame?

src/summon/browzine-summon-adapter.js
tiandavis: Intervals are capped, best to test in IT.

PR #110, opened 2022-02-04
feature/retraction-watch-overrides-show-format-choice-BZ-7024

tiandavis:
## Summary - [BZ-7024](https://thirdiron.atlassian.net/browse/BZ-7024)

Retraction Watch should override showFormatChoice.

## Description

TODO
- Adds failing test where retraction watch does not override showFormatChoice (**DONE**)
- Update implementation so retraction watch overrides showFormatChoice (**DONE**)

## Deploy Precautions

- None

### DIFF
--- a/tests/acceptance/browzine-primo-adapter-test.js
+++ b/tests/acceptance/browzine-primo-adapter-test.js
@@ -544,6 +544,142 @@ describe("BrowZine Primo Adapter >", function() {
       });
     });
 
+    describe("search results article with both retracted article link and article link >", function() {
+      beforeEach(function() {
+        primo = browzine.primo;
+        browzine.showFormatChoice = true;
+
+        searchResult = $("<div class='list-item-wrapper'><prm-brief-result-container><div class='result-item-image'><prm-search-result-thumbnail-container><img class='main-img fan-img-1' src=''/><img class='main-img fan-img-2' src=''/><img class='main-img fan-img-3' src=''/></prm-search-result-thumbnail-container></div><div class='result-item-text'><prm-search-result-availability-line><div class='layout-align-start-start'></div></prm-search-result-availability-line></div></prm-brief-result-container></div>");
+
+        inject(function ($compile, $rootScope) {
+          $scope = $rootScope.$new();
+
+          $scope.$ctrl = {
+            parentCtrl: {
+              result: {
+                pnx: {
+                  display: {
+                    type: ["article"]
+                  },
+
+                  addata: {
+                    issn: ["0028-4793"],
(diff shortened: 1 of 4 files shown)


tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -544,6 +544,142 @@ describe("BrowZine Primo Adapter >", function() {
...
+      });
+
+      it("should have an enhanced browse article in browzine option", function() {
+        var template = searchResult.find(".browzine");
+
+        expect(template).toBeDefined();
+
+        expect(template.text().trim()).toContain("View Issue Contents");
+        expect(template.text().trim()).toContain("Retracted Article");
+        expect(template.text().trim()).not.toContain("Read Article");
```
karlbecker, line 646: 👍 

Do you think it'd be useful to mention it should not contain the article link in the `it` statement? That seems to be the heart of this change. 

Edit: after looking at the Summon test, I see the language in the it used there

> should have an enhanced browse article showing retraction watch only

Think the same phrase could be used here, since they're both testing for only showing retraction watch?

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -544,6 +544,142 @@ describe("BrowZine Primo Adapter >", function() {
...
+
+        expect(template.find("a.browzine-web-link").attr("href")).toEqual("https://develop.browzine.com/libraries/XXXX/journals/8645/issues/8065238?showArticleInContext=doi:10.1186%2Fs11671-016-1523-5&utm_source=api_716");
```
karlbecker, line 648: The assertions here and below in this test seem not as important as the 3 assertions above,  but I'm guessing you're keeping these assertions in as a defense against unintentional behavior creeping into this scenario in any other part of the display we make in Primo?  

Keeping the assertions pretty tight seems fine to me 👍

karlbecker (approved):
Thanks for the cleanup on this aspect of Retraction Watch, @tiandavis - looking good! 🙌 

Just had a little question about potentially keeping the `it` string the same across both these tests, but wondering if that would be desirable or not.

Other than that - :shipit:

tests/acceptance/browzine-primo-adapter-test.js
tiandavis: Good catch @karlbecker! Definitely missed that update.

tests/acceptance/browzine-primo-adapter-test.js
tiandavis: > defense against unintentional behavior creeping into this scenario

Correctamundo!

PR #111, opened 2022-02-07
feature/retraction-watch-enhancements-BZ-7023

tiandavis:
## Summary - [BZ-7023](https://thirdiron.atlassian.net/browse/BZ-7023)

Retraction notices should be shown in discovery systems when there’s only an article link. Retraction notices should be shown in discovery systems when there is Unpaywall data. Retraction notices should be shown in discovery systems, even if something is not within a library’s holdings.

## Description

TODO:
- Retraction notices should be shown in discovery systems when there’s only an article link (**DONE**)
- Retraction notices should be shown in discovery systems when there is Unpaywall data. (**CUSTOMER HAS OLD, SELF-HOSTED, HEAVILY CUSTOMIZED PRIMO ADAPTER**)
- Retraction notices should be shown in discovery systems, even if something is not within a library’s holdings. (**DONE**) 
  - Karl referenced this as a feature enhancement in https://thirdiron.atlassian.net/browse/BZ-7036

## Deploy Precautions

- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -554,6 +554,30 @@ browzine.primo = (function() {
     return template;
   };
 
+  function retractionLinkTemplate(articleRetractionUrl) {
+    var articleLinkUrl = articleRetractionUrl;
+    var linkIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+    var linkIconWidth = "15";
+    var linkIconMarginRight = "1.5px";
+    var articleLinkText = browzine.articleRetractionWatchText || "Retracted Article";
+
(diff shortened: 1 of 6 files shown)


karlbecker (approved):
Looking great @tiandavis ! 👍  

Sorry it took me a little bit to review, but I wanted to make sure I understood all the changes and investigation done for this one since I hadn't read much of the Jira or Slack chat before looking at this code.  

Basically we went had 3 things originally identified in this Jira as changes, and: 
* one was indeed an undesired behavior we changed here
* one was a customer not realizing they forked our code 
* one was a new feature

I modified some acceptance criteria and made a new Jira to capture the new feature and separate that out from the bug report.  

Everything looks solid to me, thanks for all the investigation and tracking down that forked version, too! 🙌

PR #115, opened 2022-02-24
feature/primo-alma-delivery-category-empty-bug-BZ-7066

tiandavis:
## Summary - [BZ-7066](https://thirdiron.atlassian.net/browse/BZ-7066)

Should not filter print records when deliveryCategory is unknown

## Deploy Precautions

- None


karlbecker (approved):
Great job on the fast fix, @tiandavis ! 🙌 🏆  It was great to get the help on this from outside, too 😊 

Thanks for both permutations of that automated test, too, to help protect against either of those scenarios.

PR #119, opened 2022-07-15
feature/primo-angular-v1.8-upgrade-BZ-7188

tiandavis:
## Summary - [BZ-7188](https://thirdiron.atlassian.net/browse/BZ-7188)

Updates Primo scope extraction to handle Primo object model for both Primo Angular v1.6 and Primo Angular v1.8

## Description

Seems the object model will be slightly refactored in the next platform major release. The Angular upgrade itself may be a red herring, with the actual breakage come from an object change in the next major release. Basically, in the current Primo we could access data from two object paths. In the next major Angular release, it looks like one of those paths, the path we initially used, will be removed.

TODO:
- Fix Primo Unit and Acceptance Tests (**DONE**)

## Deploy Precautions

- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -739,7 +739,7 @@ browzine.primo = (function() {
   };
 
   function getScope($scope) {
-    return $scope && $scope.$ctrl && $scope.$ctrl.parentCtrl ? $scope.$ctrl.parentCtrl : undefined;
+    return $scope && $scope.$parent && $scope.$parent.$ctrl ? $scope.$parent.$ctrl : undefined;
   };
 
   function shouldEnhance(scope) {
--- a/tests/acceptance/browzine-primo-adapter-test.js
+++ b/tests/acceptance/browzine-primo-adapter-test.js
@@ -428,17 +428,19 @@ describe("BrowZine Primo Adapter >", function() {
         inject(function ($compile, $rootScope) {
           $scope = $rootScope.$new();
 
-          $scope.$ctrl = {
-            parentCtrl: {
-              result: {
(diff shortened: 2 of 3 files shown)


gpeterso:
Hi @tiandavis. Is it possible to get a warning for when this will be released into production? I think that it might break our [current integration](https://github.com/UMNLibraries/primo-explore-central-package/blob/main/js/components/search/search-result/search-result-availability/browzine.service.js) because we're [explicitly binding](https://github.com/UMNLibraries/primo-explore-central-package/blob/main/js/components/search/search-result/search-result-availability/browzine.component.js) the `parentCtrl` property on the object that gets passed to `searchResult($scope)`. 

I know that there are at least a few other Primo sites that use a similar approach in order to accommodate multiple `prmSearchResultAvailabilityLineAfter` components. 

Jeff

tiandavis:
Great question @gpeterso! I had a similar concern while working through this PR. What I found was I didn't have to update the component binding.

Here's what I have, everything seems to still work so far:
```
app.controller('prmSearchResultAvailabilityLineAfterController', function($scope) {
    window.browzine.primo.searchResult($scope);
});

app.component('prmSearchResultAvailabilityLineAfter', {bindings: { parentCtrl: '<' },
    controller: 'prmSearchResultAvailabilityLineAfterController'
});
```

Does that help a bit? 

We also have an Integration Testing phase before this lands to production. We’ve got your email here so can we email you during that phase of the process to make sure this works for you?

gpeterso:
Yeah, that'd be great. Thanks @tiandavis! Much appreciated.

johnmuth (approved):
👍 so many tests :)

PR #122, opened 2022-08-05
feature/library-id-override-staging-handling-BZ-7275

tiandavis:
## Summary - [BZ-7275](https://thirdiron.atlassian.net/browse/BZ-7275)

Handles staging api endpoint override when no libraryId is specified.

## Deploy Precautions

- None

### DIFF
--- a/tests/unit/models/360-core.js
+++ b/tests/unit/models/360-core.js
@@ -173,11 +173,6 @@ describe("SerSol 360 Core Model >", function() {
       expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
     });
 
-    it("should override the libraryId on the api endpoint even when an api endpoint is specified", function() {
-      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
-      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
-    });
-
     it("should return the customer supplied api endpoint when a libraryId is not specified", function() {
       delete browzine.libraryId;
       browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
--- a/tests/unit/models/360-core.js
+++ b/tests/unit/models/360-core.js
@@ -158,6 +158,33 @@ describe("SerSol 360 Core Model >", function() {
     });
   });
 
+  describe("serSol360Core model libraryIdOverride method staging >", function() {
+    beforeEach(function() {
+      delete browzine.api;
+    });
+
+    afterEach(function() {
+      delete browzine.libraryId;
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+    });
+
+    it("should not override the libraryId on the api endpoint when the libraryId is not specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
+
+    it("should override the libraryId on the api endpoint even when an api endpoint is specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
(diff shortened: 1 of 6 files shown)


tests/unit/models/360-core.js
```diff
@@ -158,6 +158,33 @@ describe("SerSol 360 Core Model >", function() {
...
+
+    it("should not override the libraryId on the api endpoint when the libraryId is not specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
+
+    it("should override the libraryId on the api endpoint even when an api endpoint is specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
```
karlbecker: Do we need to change this test?  The `it` seems to be a scenario where `libraryId` is specified in addition to `api` - is that right?

The `afterEach` call deletes `browzine.libraryId` from the object, and there's nothing in the `it` or the `describe` that sets `libraryId`

So do we want it set here, perhaps like this?

```suggestion
    it("should override the libraryId on the api endpoint even when an api endpoint is specified", function() {
      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
      browzine.libraryId = '123Y';
      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
    });
```

tests/unit/models/primo.js
```diff
@@ -295,6 +295,33 @@ describe("Primo Model >", function() {
...
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+    });
+
+    it("should not override the libraryId on the api endpoint when the libraryId is not specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(primo.libraryIdOverride(primo.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
+
+    it("should override the libraryId on the api endpoint even when an api endpoint is specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
```
karlbecker: If my previous comment is valid, then I think we'd want a change here, too.

tests/unit/models/360-core.js
```diff
@@ -158,6 +158,28 @@ describe("SerSol 360 Core Model >", function() {
...
+    it("should not override the libraryId on the api endpoint when the libraryId is not specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
+
+    it("should return the customer supplied api endpoint when a libraryId is not specified", function() {
+      delete browzine.libraryId;
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+      expect(serSol360Core.libraryIdOverride(serSol360Core.urlRewrite(browzine.api))).toEqual("https://staging-api.thirdiron.com/public/v1/libraries/XXX");
+    });
```
karlbecker, line 180: How did we go this long without realizing this? 😂 

Great job finding this @JMalone29 !

tests/unit/models/360-core.js
```diff
@@ -158,6 +158,28 @@ describe("SerSol 360 Core Model >", function() {
...
+      delete browzine.api;
+    });
+
+    afterEach(function() {
+      delete browzine.libraryId;
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
+    });
+
+    it("should not override the libraryId on the api endpoint when the libraryId is not specified", function() {
+      browzine.api = "https://staging-api.thirdiron.com/public/v1/libraries/XXX";
```
karlbecker, line 172: In both of the new tests, the libraryId is not specified, and we're expecting the same value on the same function result.  What is different about these two `it` statements?  I'm not quite telling the difference.

karlbecker (commented):
I can see the improvement here @tiandavis - excellent! But I've got a question about the tests - what do you think?

karlbecker (approved):
LGTM @tiandavis - Ship it! :shipit:

PR #126, opened 2022-11-01
Add another location where element can appear with Angular 1.8.3 and Primo Studio - bz 7459

karlbecker:
## Summary - [BZ-7459](https://thirdiron.atlassian.net/browse/BZ-7459)

Resolves a bug on the scoped to article screen

## Description

See the Jira for more details.  This resolves a bug involving the single article screen on Angular 1.8.3 and Primo Studio.

Here's a little log showing it working:
CleanShot 2022-10-31 at 22 51 30@2x

Think there's a way we could get this under automated test @tiandavis ?

## Deploy Precautions

- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -750,7 +750,7 @@ browzine.primo = (function() {
 
       // Post-Primo Angular 1.8.3
       // 3i Normal Integration: ✗
-      // Primo Studio Integration: ✓
+      // Primo Studio Integration: ✗
       scope = $scope.$ctrl.parentCtrl;
     } else if ($scope && $scope.$parent && $scope.$parent.$ctrl && $scope.$parent.$ctrl.$element) {
       // Pre-Primo Angular 1.8.3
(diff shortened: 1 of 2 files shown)


src/primo/browzine-primo-adapter.js
```diff
@@ -761,6 +761,15 @@ browzine.primo = (function() {
       // 3i Normal Integration: ✓
       // Primo Studio Integration: ✗
       scope = $scope.$parent.$ctrl;
+    } else if ($scope && $scope.$parent && $scope.$parent.$ctrl.parentCtrl && $scope.$parent.$ctrl.parentCtrl.$element) {
```
tiandavis, line 764: Complete walking the line by adding a `$scope.$parent.$ctrl` check to make sure it exists:
$scope && $scope.$parent **&& $scope.$parent.$ctrl** && $scope.$parent.$ctrl.parentCtrl && $scope.$parent.$ctrl.parentCtrl.$element

tiandavis (changes requested):
Good analysis!

PR Feedback and we could probably use some tests to prevent regressions.

Additional Unit Test suggestion and example (something like):
`should retrieve the scope after clicking a specific search result`
https://github.com/thirdiron/browzine-discovery-service-adapters/blob/3c9a5bd68922983b19338f3703b43ee2913e6d97/tests/unit/models/primo.js#L182-L215

The acceptance tests are optimized for the search results, so there is no good way to test a "search result details" type scenario. Pretty sure the additional unit test should be enough though.

src/primo/browzine-primo-adapter.js
karlbecker: Whoops, yes, thanks for catching that Tian!

karlbecker:
Thanks for the guidance on how to do this, @tiandavis ! That was really helpful.

I gave it a crack to add a new test, making sure it failed in CI without the fix, and passed with the fix.  How does that look?

karlbecker:
One more commit I pushed up @tiandavis , I think I'm done now: I noticed since this test I added is related to an "article" type, the test should probably be looking for a DOI rather than an ISSN.  Let me know what you think - thanks!

tiandavis (approved):
@karlbecker Yep! Test looks great Karl! 🛳 🎉

PR #127, opened 2023-09-11
Adding EOC notices to Primo

Evroble:
## Summary - [BZ-7639](https://thirdiron.atlassian.net/browse/BZ-7639)

This release will add Expression of Concern notices to Primo instances, similar to the pre-existing Retraction notices.

## Description

Before EOC Notifications:
CleanShot 2023-09-26 at 12 26 10

After EOC Notifications:
CleanShot 2023-09-26 at 12 05 03

## Deploy Precautions

- None

[BZ-7639]: https://thirdiron.atlassian.net/browse/BZ-7639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -1120,7 +1120,7 @@ browzine.primo = (function() {
     showArticleLink: showArticleLink,
     showPrintRecords: showPrintRecords,
     showRetractionWatch: showRetractionWatch,
-    showExpressionOfConcern: showExpressionOfConcern;
+    showExpressionOfConcern: showExpressionOfConcern,
     showFormatChoice: showFormatChoice,
     showLinkResolverLink: showLinkResolverLink,
     enableLinkOptimizer: enableLinkOptimizer,
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -728,9 +728,7 @@ browzine.primo = (function() {
     return template;
   };
 
-  //There's a bit of a question on whether we would need a retraction notice/eoc notice on this as we'd only get to this function
-  //if we don't already have a PDF link-we might only link to these live unpaywall responses if they have a direct-to-PDF link
-  //But we may want to investigate this to be sure...
+  //TODO: BZ 8196 - Will address the need here for retraction/eoc handling as well as ironing out some logic.
   function unpaywallArticleLinkTemplate(articleLinkUrl) {
     var linkIcon = "https://assets.thirdiron.com/images/integrations/browzine-article-link-icon.svg";
     var articleLinkText = browzine.articleLinkViaUnpaywallText  || "Read Article (via Unpaywall)";
@@ -919,7 +917,7 @@ browzine.primo = (function() {
           libKeyLinkOptimizer.innerHTML += template;
         }
 
-        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && !articleEocNoticeUrl && isArticle(scope) && showRetractionWatch()) {
+        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && isArticle(scope) && showRetractionWatch()) {
           var template = retractionWatchLinkTemplate(articleRetractionUrl);
           libKeyLinkOptimizer.innerHTML += template;
         }
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -248,6 +248,7 @@ browzine.primo = (function() {
     if (!isArticle(scope)) {
       return false;
     }
+    //We can maybe get rid of this now? Per Karl, the reasons for doing this may no longer matter. May need more investigation.
     if (!data || !data.hasOwnProperty("unpaywallUsable")) {
       return true;
     }
@@ -277,6 +278,17 @@ browzine.primo = (function() {
 
     return articleRetractionUrl;
   };
+
+  function getArticleEOCNoticeUrl(scope, data) {
+    var articleEocNoticeUrl = null;
+
+    if (isArticle(scope)) {
+      if (data && data.expressionOfConcernNoticeUrl) {
+        articleEocNoticeUrl = data.expressionOfConcernNoticeUrl;
+      }
+    }
+    return articleEocNoticeUrl;
+  }
 
   function isTrustedRepository(response) {
     var validation = false;
@@ -448,6 +460,17 @@ browzine.primo = (function() {
     return featureEnabled;
   };
 
+  function showExpressionOfConcern() {
+    var featureEnabled = false;
+    var config = browzine.articleExpressionOfConcernEnabled;
+
+    if(typeof config === "undefined" || config === null || config === true) {
+      featureEnabled = true;
+    };
+
+    return featureEnabled;
+  };
+
   function showFormatChoice() {
     var featureEnabled = false;
     var config = browzine.showFormatChoice;
@@ -515,6 +538,10 @@ browzine.primo = (function() {
     return articleRetractionUrl && showRetractionWatch();
   };
 
+  function showEocNoticeUI(articleEocNoticeUrl) {
+    return articleEocNoticeUrl && showExpressionOfConcern();
+  }
+
   function directToPDFTemplate(directToPDFUrl, articleRetractionUrl) {
     var pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-pdf-download-icon.svg";
     var pdfIconWidth = "12";
@@ -527,6 +554,14 @@ browzine.primo = (function() {
       pdfIconWidth = "15";
       pdfIconMarginRight = "1.5px";
       articlePDFDownloadLinkText = browzine.articleRetractionWatchText || "Retracted Article";
+    }
+
+    if (showEocNoticeUI(articleEocNoticeUrl) && !showRetractionWatchUI(articleRetractionUrl)) {
+      directToPDFUrl = articleEocNoticeUrl;
+      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      pdfIconWidth = "15";
+      pdfIconMarginRight = "1.5px";
+      articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     }
 
     var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
@@ -558,6 +593,14 @@ browzine.primo = (function() {
       linkIconWidth = "15";
       linkIconMarginRight = "1.5px";
       articleLinkText = browzine.articleRetractionWatchText || "Retracted Article";
+    }
+
+    if (showEocNoticeUI(articleEocNoticeUrl) && !showRetractionWatchUI(articleRetractionUrl)) {
+      articleLinkUrl = articleEocNoticeUrl;
+      linkIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      linkIconWidth = "15";
+      linkIconMarginRight = "1.5px";
+      articleLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     }
 
     var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
@@ -601,6 +644,31 @@ browzine.primo = (function() {
     return template;
   };
 
+  function eocLinkTemplate(articleEocNoticeUrl) {
+    var articleLinkUrl = articleEocNoticeUrl;
+    var linkIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+    var linkIconWidth = "15";
+    var linkIconMarginRight = "1.5px";
+    var articleLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
+
+    var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
+                      "<a class='browzine-article-link' href='{articleLinkUrl}' target='_blank' onclick='browzine.primo.transition(event, this)'>" +
+                          "<img alt='BrowZine Article Link Icon' src='{linkIcon}' class='browzine-article-link-icon' style='margin-bottom: -3px; margin-right: {linkIconMarginRight};' aria-hidden='true' width='{linkIconWidth}' height='16'/> " +
+                          "<span class='browzine-article-link-text'>{articleLinkText}</span> " +
+                          "<md-icon md-svg-icon='primo-ui:open-in-new' class='md-primoExplore-theme' aria-hidden='true' style='height: 15px; width: 15px; min-height: 15px; min-width: 15px; margin-top: -2px;'><svg width='100%' height='100%' viewBox='0 0 24 24' y='504' xmlns='http://www.w3.org/2000/svg' fit='' preserveAspectRatio='xMidYMid meet' focusable='false'><path d='M14,3V5H17.59L7.76,14.83L9.17,16.24L19,6.41V10H21V3M19,19H5V5H12V3H5C3.89,3 3,3.9 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V12H19V19Z'></path></svg></md-icon>" +
+                      "</a>" +
+                   "</div>";
+
+    template = template.replace(/{articleLinkUrl}/g, articleLinkUrl);
+    template = template.replace(/{articleLinkText}/g, articleLinkText);
+    template = template.replace(/{linkIcon}/g, linkIcon);
+    template = template.replace(/{linkIconWidth}/g, linkIconWidth);
+    template = template.replace(/{linkIconMarginRight}/g, linkIconMarginRight);
+
+    return template;
+  };
+
+
   function browzineWebLinkTemplate(scope, browzineWebLink) {
     var browzineWebLinkText = "";
     var bookIcon = "https://assets.thirdiron.com/images/integrations/browzine-open-book-icon.svg";
@@ -628,7 +696,7 @@ browzine.primo = (function() {
     return template;
   };
 
-  function unpaywallArticlePDFTemplate(directToPDFUrl, articleRetractionUrl) {
+  function unpaywallArticlePDFTemplate(directToPDFUrl, articleRetractionUrl, articleEocNoticeUrl) {
     var pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-pdf-download-icon.svg";
     var pdfIconWidth = "12";
     var pdfIconMarginRight = "4.5px";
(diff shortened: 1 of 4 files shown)


src/primo/browzine-primo-adapter.js
```diff
@@ -248,6 +248,7 @@ browzine.primo = (function() {
     if (!isArticle(scope)) {
       return false;
     }
+    //We can maybe get rid of this now? Per Karl, the reasons for doing this may no longer matter. May need more investigation.
```
karlbecker, line 251: I think it's fine to keep this and eliminate the comment, since this will keep on working fine without any change.  But if we want to eliminate it, sure, we can axe it and see how it behaves.  But I think in this case, since we don't change this too often, and don't mess with unpaywall too much, we can probably just keep it the same. This is a pretty defensive approach to this code, since it checks to see if the API response even has this element, and if it doesn't does the right thing.  If we ever get rid of that element, it would be nice to know we have defensive code deployed that likely won't break.

src/primo/browzine-primo-adapter.js
```diff
@@ -278,6 +279,18 @@ browzine.primo = (function() {
     return articleRetractionUrl;
   };
 
+  function getArticleEOCNoticeUrl(scope, data) {
+    var articleEocNoticeUrl = null;
+
+    if (isArticle(scope)) {
+      if (data && data.expressionOfConcernNoticeUrl) {
+        console.log(data.expressionOfConcernNoticeUrl, 'here is our expressionOfConcernNoticeUrl');
```
karlbecker, line 287: We can delete 🔥  this console log now that we're getting ready to merge this. Probably very helpful during debugging though 😊 

```suggestion
```

src/primo/browzine-primo-adapter.js
```diff
@@ -448,6 +461,17 @@ browzine.primo = (function() {
     return featureEnabled;
   };
 
+  function showExpressionOfConcern() {
+    var featureEnabled = false;
+    var config = browzine.articleExpressionOfConcernEnabled;
```
karlbecker, line 466: Now that we have a new configuration option, you'll want to check with either JohnS or Nick about where they want to add it, if anywhere.

At a minimum, https://thirdiron.atlassian.net/wiki/spaces/BrowZineAPIDocs/pages/79200260/Ex+Libris+Primo+Integration and https://thirdiron.atlassian.net/wiki/spaces/BrowZineAPIDocs/pages/104988678/Ex+Libris+Summon+Overview+and+Step-by-Step+Installation+Guide in the various Options (look for the expandable sections, they have the configs thoroughly specified there)

src/primo/browzine-primo-adapter.js
```diff
@@ -529,6 +557,14 @@ browzine.primo = (function() {
       articlePDFDownloadLinkText = browzine.articleRetractionWatchText || "Retracted Article";
     }
 
+    if (showEocNoticeUI(articleEocNoticeUrl) && !showRetractionWatchUI(articleRetractionUrl)) {
```
karlbecker, line 560: One thought: if this shouldn't happen when showRetractionWatchUI is truthy, there's at least another way we could change this to reduce a little code here:

```suggestion
    } else if (showEocNoticeUI(articleEocNoticeUrl)) {
```

By doing this, if we have `showRetractionWatchUI(articleRetractionUrl)` truthy, then the stuff on the above lines runs, but if that's falsy, then we give the EocNotice its attempt.

src/primo/browzine-primo-adapter.js
```diff
@@ -529,6 +557,14 @@ browzine.primo = (function() {
...
+      directToPDFUrl = articleEocNoticeUrl;
+      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      pdfIconWidth = "15";
+      pdfIconMarginRight = "1.5px";
+      articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
```
karlbecker, line 565: Same with this for a config option - you'll want to tell someone in ops about potentially adding this to publicly-facing config documentation, too.

Although this GitHub repo is publicly accessible, since most people don't need to look at the code, most of the documentation integrators use are the Confluence ones.

src/primo/browzine-primo-adapter.js
```diff
@@ -560,6 +596,14 @@ browzine.primo = (function() {
       articleLinkText = browzine.articleRetractionWatchText || "Retracted Article";
     }
 
+    if (showEocNoticeUI(articleEocNoticeUrl) && !showRetractionWatchUI(articleRetractionUrl)) {
```
karlbecker, line 599: Same potential here to eliminate one part of the `if` clause.

src/primo/browzine-primo-adapter.js
```diff
@@ -601,6 +645,31 @@ browzine.primo = (function() {
...
+
+    var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
+                      "<a class='browzine-article-link' href='{articleLinkUrl}' target='_blank' onclick='browzine.primo.transition(event, this)'>" +
+                          "<img alt='BrowZine Article Link Icon' src='{linkIcon}' class='browzine-article-link-icon' style='margin-bottom: -3px; margin-right: {linkIconMarginRight};' aria-hidden='true' width='{linkIconWidth}' height='16'/> " +
+                          "<span class='browzine-article-link-text'>{articleLinkText}</span> " +
+                          "<md-icon md-svg-icon='primo-ui:open-in-new' class='md-primoExplore-theme' aria-hidden='true' style='height: 15px; width: 15px; min-height: 15px; min-width: 15px; margin-top: -2px;'><svg width='100%' height='100%' viewBox='0 0 24 24' y='504' xmlns='http://www.w3.org/2000/svg' fit='' preserveAspectRatio='xMidYMid meet' focusable='false'><path d='M14,3V5H17.59L7.76,14.83L9.17,16.24L19,6.41V10H21V3M19,19H5V5H12V3H5C3.89,3 3,3.9 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V12H19V19Z'></path></svg></md-icon>" +
+                      "</a>" +
+                   "</div>";
```
karlbecker, line 661: Do you have a screenshot of what this looks like in Primo? Seeing a screenshot in the PR description would be pretty nice to give a quick confirmation that it's generally looking how we'd expect.  Otherwise I'm really squinting at this HTML and trying to imagine what it looks like 😊

src/primo/browzine-primo-adapter.js
```diff
@@ -659,6 +736,9 @@ browzine.primo = (function() {
     return template;
   };
 
+  //There's a bit of a question on whether we would need a retraction notice/eoc notice on this as we'd only get to this function
+  //if we don't already have a PDF link-we might only link to these live unpaywall responses if they have a direct-to-PDF link
+  //But we may want to investigate this to be sure...
```
karlbecker, line 741: Unless you're thinking of investigating it, I think we can remove this comment... did you find an example of when this would get triggered?

You could ask Jay or Nick to see if they have an example of an article in Primo or Summon that would trigger a live Unpaywall call.  Then you could mark it as having an EOC and see how it behaves.  

If we want to investigate this later, let's create a new Jira describing what we're investigating and what question(s) we need to answer, and put the Jira number as a prefix to this comment like `TODO BZ-1234 - ` .  But if it seems like not a big deal, we can just remove this comment.

src/primo/browzine-primo-adapter.js
```diff
@@ -692,6 +772,14 @@ browzine.primo = (function() {
       articlePDFDownloadLinkText = browzine.articleRetractionWatchText || "Retracted Article";
     }
 
+    if (showEocNoticeUI(articleEocNoticeUrl) && !showRetractionWatchUI(articleRetractionUrl)) {
```
karlbecker, line 775: Also here, can use an else if

src/primo/browzine-primo-adapter.js
```diff
@@ -831,20 +920,25 @@ browzine.primo = (function() {
...
           libKeyLinkOptimizer.innerHTML += template;
         }
 
-        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && isArticle(scope) && showRetractionWatch()) {
+        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && !articleEocNoticeUrl && isArticle(scope) && showRetractionWatch()) {
           var template = retractionWatchLinkTemplate(articleRetractionUrl);
           libKeyLinkOptimizer.innerHTML += template;
         }
 
+        if (!directToPDFUrl && !articleLinkUrl && !articleRetractionUrl && articleEocNoticeUrl && isArticle(scope) && showExpressionOfConcern()) {
```
karlbecker, line 937: It's a long if statement, but this is what we want! 👍

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -554,6 +554,141 @@ describe("BrowZine Primo Adapter >", function() {
...
+
+        expect(request.url).toMatch(/articles\/doi\/10.1001%2Fjama.298.4.413/);
+        expect(request.method).toBe('GET');
+      });
+
+      afterEach(function() {
+        jasmine.Ajax.uninstall();
+      });
+
+      //Need to edit the following tests to be about EOC rather than Retraction
```
karlbecker: As long as you've fully done this, you can remove this comment 😊 

```suggestion
```

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -554,6 +554,141 @@ describe("BrowZine Primo Adapter >", function() {
...
+          });
+
+          done();
+        });
+      });
+
+      it("should open a new window when an eoc article link is clicked", function() {
+        spyOn(window, "open");
+        searchResult.find(".browzine .browzine-direct-to-pdf-link").click();
+        expect(window.open).toHaveBeenCalledWith("https://develop.libkey.io/libraries/XXXX/10.1001/jama.298.4.413", "_blank");
```
karlbecker, line 680: 👍

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -554,6 +554,141 @@ describe("BrowZine Primo Adapter >", function() {
...
+
+      //Need to edit the following tests to be about EOC rather than Retraction
+      it("should have an enhanced browse article in browzine option", function() {
+        var template = searchResult.find(".browzine");
+
+        expect(template).toBeDefined();
+        console.log(template.text(), 'this is our template text');
+
+        expect(template.text().trim()).toContain("View Issue Contents");
+        expect(template.text().trim()).toContain("Expression of Concern");
```
karlbecker, line 653: 👍

tests/unit/models/primo.js
```diff
@@ -2236,6 +2386,67 @@ describe("Primo Model >", function() {
...
+
+      expect(template).toContain("Expression of Concern");
+      expect(template).toContain("https://develop.libkey.io/libraries/XXXX/10.1002/ijc.25451");
+      expect(template).toContain("https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg");
+    });
+
+    it("should apply the articleExpressionOfConcernText config property when eoc notice available and eoc enabled", function() {
+      browzine.articleExpressionOfConcernText = "Expression of Concern PDF";
```
karlbecker, line 2425: Since this link will never go directly to a PDF, but instead to LibKey Web, what if we changed the test text here to something a little different?  It can certainly be goofy 😊 

```suggestion
      browzine.articleExpressionOfConcernText = "I have a concern that must be expressed!";
```

tests/unit/models/primo.js
```diff
@@ -2236,6 +2386,67 @@ describe("Primo Model >", function() {
...
+        }
+      };
+
+      var data = primo.getData(eocArticleResponse);
+      var directToPDFUrl = primo.getDirectToPDFUrl(scope, data);
+      var articleRetractionUrl = primo.getArticleRetractionUrl(scope, data);
+      var articleEOCUrl = primo.getArticleEOCNoticeUrl(scope, data);
+      var template = primo.directToPDFTemplate(directToPDFUrl, articleRetractionUrl, articleEOCUrl);
+
+      expect(template).toContain("Expression of Concern PDF");
```
karlbecker, line 2448: And then switch it down here, too:

```suggestion
      expect(template).toContain("I have a concern that must be expressed!");
```

The reason I like making it a little goofier and not mention PDF is because I don't want any questions to come up in the future when others are reviewing this and wondering, "Hmmm, so does this go to a PDF file? This test seems to indicate it probably does."

So making it something very different will hopefully help eliminate that question from coming up in people's minds.

tests/unit/models/primo.js
```diff
@@ -2376,6 +2587,75 @@ describe("Primo Model >", function() {
...
+
+      expect(template).toContain("Expression of Concern");
+      expect(template).toContain("https://develop.libkey.io/libraries/XXXX/10.1002/ijc.25451");
+      expect(template).toContain("https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg");
+    });
+
+    it("should apply the articleExpressionOfConcernText config property", function() {
+      browzine.articleExpressionOfConcernText = "Expression of Concern (Proceed with Caution)";
```
karlbecker, line 2634: Bingo, like this one 👍 nice.

tests/unit/models/primo.js
```diff
@@ -23,7 +23,18 @@ describe("Primo Model >", function() {
...
+        },
+        {
+          "id": 13016,
+          "type": "journals",
+          "title": "International Journal of Cancer",
+          "issn": "00207136",
+          "sjrValue": 2.259,
+          "coverImageUrl": "https://assets.thirdiron.com/images/covers/0020-7136.png",
+          "browzineEnabled": true,
+          "browzineWebLink": "https://develop.browzine.com/libraries/XXXX/journals/13016?utm_source=api_716"
```
karlbecker, line 35: Not sure I fully follow why we added a new journal here.  Could we have just added a new article that is associated with one of the other journals, but that has an expressionOfConcernNoticeUrl ?

karlbecker (commented):
Hey @Evroble , this is looking pretty good - nice job understanding the requirements, following the patterns in this repo that were already setup in the past, adding many relevant tests (and following the pattern for adding tests, too), and documenting some of the questions we had, too.

I've got a couple small changes I think we could do, although not all of them I'm 100% sure on what we should do. For instance, for the comments where you said we could investigate more, we could make a TODO and do that, or we could ask folks now for a little more info and then see if it's a concern.

Let me know what you think - but this is really close to merging I think, it could just use a couple small tweaks here and there. 

Great effort on your first Discovery Service Adapter change!  🙌

src/primo/browzine-primo-adapter.js
Evroble: @karlbecker Let me ping Jay and Nick and see. So essentially we need an example where we have a retracted or EOC article that doesn't have a pdf or article url, doesn't have an unpaywall pdf? I'm trying to think of how to formulate the question but admittedly I don't fully understand the flow of when we'd get to this function.

tests/unit/models/primo.js
```diff
@@ -2445,7 +2430,7 @@ describe("Primo Model >", function() {
       var articleEOCUrl = primo.getArticleEOCNoticeUrl(scope, data);
       var template = primo.directToPDFTemplate(directToPDFUrl, articleRetractionUrl, articleEOCUrl);
 
-      expect(template).toContain("Expression of Concern PDF");
+      expect(template).toContain("An area of grave concern");
```
karlbecker, line 2433: 💀

karlbecker (approved):
Looks good to me @Evroble 👍

src/primo/browzine-primo-adapter.js
karlbecker: @Evroble - I think that's right, yeah.  Quite a corner case. I wouldn't even ask about the details, I'd just ask if they know of any articles that require an Unpaywall call to show data.  If they can find one or two, then we can always mark it as EOC in the staging database for a little bit.

src/primo/browzine-primo-adapter.js
```diff
@@ -831,20 +910,25 @@ browzine.primo = (function() {
...
 
-        if ((!directToPDFUrl || (showFormatChoice() && !articleRetractionUrl)) && articleLinkUrl && isArticle(scope) && showDirectToPDFLink() && showArticleLink()) {
-          var template = articleLinkTemplate(articleLinkUrl, articleRetractionUrl);
+        if ((!directToPDFUrl || (showFormatChoice() && !articleRetractionUrl && !articleEocNoticeUrl)) && articleLinkUrl && isArticle(scope) && showDirectToPDFLink() && showArticleLink()) {
+          var template = articleLinkTemplate(articleLinkUrl, articleRetractionUrl, articleEocNoticeUrl);
           libKeyLinkOptimizer.innerHTML += template;
         }
 
-        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && isArticle(scope) && showRetractionWatch()) {
+        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && !articleEocNoticeUrl && isArticle(scope) && showRetractionWatch()) {
```
karlbecker, line 922: Let's modify this to be like this:

```suggestion
        if (!directToPDFUrl && !articleLinkUrl && articleRetractionUrl && isArticle(scope) && showRetractionWatch()) {
```

It's an unlikely scenario, but even if an article has both an EOC and a retraction, we want to show the retraction data.

src/primo/browzine-primo-adapter.js
karlbecker: What an investigation @Evroble and I did! We had to dive into the past, but here's a tl;dr: 

We identified that there's a bug that probably isn't triggered too often in the code... and we don't even have an article handy we can test it with.

Here's what we found:
1. If an API call is made, and that response has neither an article link nor a directToPdfUrl, but it does have a retracted flag on it, we will insert a retracted article link into search results thanks to this line of code that's been in there a long time: https://github.com/thirdiron/browzine-discovery-service-adapters/pull/127/files#diff-e538aa7cc112daa1deb5305ef7f4b9feda2970ad04ec12f05dc5b2dc2f9b05edR923
2. After that, the code will continue on to this line: `if ((request.readyState == XMLHttpRequest.DONE && request.status == 404) || (isArticle(scope) && (!directToPDFUrl && !articleLinkUrl && unpaywallUsable))) {`
3. Since that same article does not have direct to PDF or article link, it'll enter that if statement, and then if Unpaywall knows about it, we'll also make a new button there if unpaywall has a PDF
    * That's because the `unpaywallArticlePDFTemplate` and `unpaywallManuscriptPDFTemplate` functions get called, and they insert a link themselves at this point

So we should clean up this logic so only one or the other happens.

Great questions you asked Emily! 🙌

karlbecker (changes requested):
One tiny change from our chat @Evroble , but once that's in place I think we're good to go - and we can add a TODO and its associated Jira for the bug we found, too.

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -518,6 +519,7 @@ describe("BrowZine Primo Adapter >", function() {
 
         expect(template.text().trim()).toContain("View Issue Contents");
         expect(template.text().trim()).toContain("Retracted Article");
+        expect(template.text().trim()).not.toContain("Expression of Concern");
```
karlbecker, line 522: Nice improvement 👍

karlbecker (approved):
LGTM @Evroble 🎉

PR #128, opened 2023-09-12
Remove the unused discoveryServiceBehavior value

karlbecker:
## Summary

Remove a value from the tests that has already been removed from the API in the CMS repo

## Description

We removed it from the public API as part of BZ-7503, and it was only ever potentially used by this repo, so let's get rid of it from the fixture data in this repo. 

We ended up not actually using this, since our strategy changed from the "onelink" approach that this setting was intended to express.

## Deploy Precautions

- None


Evroble (approved):
LGTM 👍

PR #132, opened 2024-05-03
Discovery Service Adapters do not call unpaywall when unpaywallUsable is false

Evroble:
## Summary - [BZ-8610](https://thirdiron.atlassian.net/browse/BZ-8610)

This change allows the discovery service adapters to interpret the avoidUnpaywall variable that the TIAPI sends on some 404 returns.

## Deploy Precautions

- :warning: Depends on another PR getting into production first: CMS change must be deployed first - https://github.com/thirdiron/browzine_cms/pull/4289

[BZ-8610]: https://thirdiron.atlassian.net/browse/BZ-8610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -872,7 +872,7 @@ browzine.primo = (function() {
   };
 
   function shouldUnpaywallLiveCallWhen404(response) {
-    if (response.hasOwnProperty('meta')&& response.meta.hasOwnProperty('unpaywallUsable')) {
+    if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('unpaywallUsable')) {
       return response.meta.unpaywallUsable;
     } else {
       return true;
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -871,11 +871,11 @@ browzine.primo = (function() {
     return validation;
   };
 
-  function shouldUnpaywallLiveCallWhen404(response) {
-    if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('unpaywallUsable')) {
-      return response.meta.unpaywallUsable;
+  function shouldAvoidUnpaywall(response) {
+    if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('avoidUnpaywall')) {
+      return response.meta.avoidUnpaywall;
     } else {
-      return true;
+      return false;
     };
   }
 
@@ -995,8 +995,8 @@ browzine.primo = (function() {
       if ((request.readyState == XMLHttpRequest.DONE && request.status == 404) || (isArticle(scope) && (!directToPDFUrl && !articleLinkUrl && unpaywallUsable))) {
 
         var response = JSON.parse(request.response || '{}');
-        var shouldUnpaywallLiveCall = shouldUnpaywallLiveCallWhen404(response);
-        if (!shouldUnpaywallLiveCall) {
+        var shouldAvoidUnpaywallLiveCall = shouldAvoidUnpaywall(response);
+        if (shouldAvoidUnpaywallLiveCall) {
           return
         }
 
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -872,8 +872,8 @@ browzine.primo = (function() {
   };
 
   function shouldUnpaywallLiveCallWhen404(response) {
-    if (response.hasOwnProperty('unpaywallUsable')) {
-      return response.unpaywallUsable;
+    if (response.hasOwnProperty('meta')&& response.meta.hasOwnProperty('unpaywallUsable')) {
+      return response.meta.unpaywallUsable;
     } else {
       return true;
     };
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -121,7 +121,7 @@ browzine.primo = (function() {
 
     if (isArticle(scope) && getDoi(scope)) {
       var doi = getDoi(scope);
-      endpoint = api + "/articles/doi/" + doi + "?include=journal,library";
+      endpoint = api + "/articles/doi/" + doi + "?include=journal,library&include_suppressed=true";
     }
 
     if (isArticle(scope) && !getDoi(scope) && getIssn(scope)) {
@@ -871,6 +871,14 @@ browzine.primo = (function() {
     return validation;
   };
 
+  function shouldUnpaywallLiveCallWhen404(response) {
+    if (response.hasOwnProperty('unpaywallUsable')) {
+      return response.unpaywallUsable;
+    } else {
+      return true;
+    };
+  }
+
   function searchResult($scope) {
     var scope = getScope($scope);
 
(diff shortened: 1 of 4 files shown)


src/summon/browzine-summon-adapter.js
karlbecker, line 1: I don't see any `include_suppressed=true` in this file - if we're not including it, then something is amiss.

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -3412,5 +3412,60 @@ describe("BrowZine Summon Adapter >", function() {
...
+      afterEach(function() {
+        delete browzine.unpaywallEmailAddressKey;
+        delete browzine.articlePDFDownloadViaUnpaywallEnabled;
+        delete browzine.articleLinkViaUnpaywallEnabled;
+        delete browzine.articleAcceptedManuscriptPDFViaUnpaywallEnabled;
+        delete browzine.articleAcceptedManuscriptArticleLinkViaUnpaywallEnabled;
+
+        jasmine.Ajax.uninstall();
+      });
+      it("Does not call unpaywall when unpaywallUsable=false", function () {
```
karlbecker, line 3462: We're not checking for the flag in the `thirdIronApiDoiRequest.respondWith` and therefore that's why this test isn't failing... and how it will never fail even if we don't put `include_suppressed`

Let's add a check both here and in the Primo test to ensure `include_suppressed` is included on the API call to the Third Iron API.

This example from [the docs](https://github.com/jasmine/jasmine-ajax?tab=readme-ov-file#5-inspect-ajax-requests) should be similar to what we need:

```
    request = jasmine.Ajax.requests.mostRecent();
    expect(request.url).toBe('venues/search');
    expect(request.method).toBe('POST');
    expect(request.data()).toEqual({latLng: ['40.019461, -105.273296']});
```

We can check for the request.url to contain `include_suppressed=true`

karlbecker (changes requested):
Hey @Evroble , I think we've at least got a bug. The Summon adapter is not adding the `include_suppressed` query parameter as far as I can tell ☺️  and that means things won't work right.

I've included a way to both improve the tests (both Primo and Summon tests) to ensure they check for that query param to be there.  Once those checks are there in the tests, I think the Summon test should fail, and then you can add a commit to fix Summon's behavior.

tests/acceptance/browzine-summon-adapter-test.js
Evroble: To clarify, are you saying we need to add an expect, to ensure the api call contains ```include_suppressed=true``` AND also add something onto our ```thirdIronApiDoiRequest.respondWith```?

src/summon/browzine-summon-adapter.js
mike-lang: That said, `include_suppressed` feels to me like a weird thing for us to expose on our public API (especially if it still returns a 404 in certain suppressed cases).  What do you all think about just returning the unpaywall live calls switch on all 404s where both the article is suppressed and the unpaywall data is supressed?

Evroble: @mike-lang I think the difficulty is that currently with the public API route, if we do not include "include suppressed" we have no way of knowing why an article is returning a 404 since it won't return any data. In the CMS we could just always tell the code to include_suppressed, but Karl was concerned that since that public route is used by other systems, that it would be confusing if they suddenly started getting "unpaywallUsable:false". We could rename "include_suppressed" to something more vague, and have the public API code interpret that value if you think that would be a better option. But i'm open to other suggestions too.

karlbecker: @Evroble I think this was already addressed over in the backend PR, but in response to...

>  we have no way of knowing why an article is returning a 404 since it won't return any data

...the idea is to still include the `unpaywallUsable: false` bit in the response on the public API with a 404. 

> Karl was concerned that since that public route is used by other systems, that it would be confusing if they suddenly started getting "unpaywallUsable:false".

Indeed I was, but I'm flexible 😊  Since @mike-lang is feeling fine about it, yeah, let's simplify the design so we don't need the query parameter. If any third party really gets confused, we can always add the query parameter later. 

> What do you all think about just returning the unpaywall live calls switch on all 404s where both the article is suppressed and the unpaywall data is supressed?

Works for me if you're pretty confident it won't cause issues @mike-lang 👍

tests/acceptance/browzine-summon-adapter-test.js
karlbecker: Hey @Evroble , 
>  are you saying we need to add an expect, to ensure the api call contains include_suppressed=true 

Yes 👍 but since it sounds like we'll change the API so it's not needed, then no worries, we just eliminate that parameter and the checks in the tests.

> AND also add something onto our thirdIronApiDoiRequest.respondWith?

Nope, nothing additional to add there.

karlbecker: However @Evroble , one recommendation for the future: you put both the test changes and the behavior change in the same function:

CleanShot 2024-05-07 at 21 27 43@2x

To help with the code review, I'd suggest putting the test that you expect to fail in its own commit. By keeping that separate from the logic that will get the test to pass, you can prove both to yourself and to a reviewer that the test you added properly fails when the necessary logic is missing. 

So two commits next time: one to modify just the test and show it's failing in CI, followed by another commit that shows you got the tests go to green.

I'm trying to find an article that discusses this method, but I'm surprisingly coming up short.  This is the closest one I can find: https://www.ranorex.com/blog/test-driven-development/ 

That article says:

> And the melancholy reality is that Red always remains on the periphery of our routines, because we don’t share Red with our peers. 

And that's what I'm recommending to avoid.  Instead, we should indeed share those failing commits in the feature branch, so everyone can see the test fails as expected. Otherwise, if the test doesn't fail in CI, then we have a test that isn't truly capturing the requirement.

Hope that's clear, happy to chat about it if you have any questions 😊

src/primo/browzine-primo-adapter.js
```diff
@@ -871,6 +871,14 @@ browzine.primo = (function() {
     return validation;
   };
 
+  function shouldAvoidUnpaywall(response) {
+    if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('avoidUnpaywall')) {
```
mike-lang, line 875: Shouldn't this use the value of `response.meta.avoidUnpaywall` rather than just check for the presence of a value?  What if someone made a CMS change that make the response have an `avoidUnpaywall` value of false for some reason?  With this code, a value of false would make the discovery service adapter avoid making an unpaywall call.

src/summon/browzine-summon-adapter.js
```diff
@@ -886,6 +886,14 @@ browzine.summon = (function() {
     return primaryColor;
   };
 
+  function shouldAvoidUnpaywall(response) {
+    if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('avoidUnpaywall')) {
```
mike-lang, line 890: Same comment in regards to `hasOwnProperty` here as the other adapter.

mike-lang (changes requested):
This mostly looks good, but I think we should use the actual value of `response.meta.avoidUnpaywall` rather than use just the presence of the property.  Change that and I think we're good.

src/primo/browzine-primo-adapter.js
Evroble: I'm not sure I follow. This function returns the value of avoidUnpaywall and then on line 999 we interpret that value so that if it's true, we won't make an unpaywall call, but if its false we will. So in the scenario you mentioned, if avoidUnpaywall = false, then we'd make an unpaywall call.

mike-lang: Oh nevermind, I was just thrown off by the use of `hasOwnProperty`.  This whole function could be replaced with 
```
return !!(response.meta && response.meta.avoidUnpaywall);
```
Either way looks good to me now.

Evroble: I find the way i've written this part to be easier to understand at a glance, so if that's okay with you then are we good to merge?

mike-lang: Yep this is fine to merge.

PR #134, opened 2024-06-24
Update shouldAvoidUnpaywall to avoid publisher links

Evroble:
## Summary - [BZ-8727](https://thirdiron.atlassian.net/browse/BZ-8727)

This update makes it so Primo and Summons will avoid using Unpaywall Publisher Links when directed.

## Description

BEFORE"

AFTER:
CleanShot 2024-06-24 at 16 16 01

## Implementation Notes

1. ​

## Deploy Precautions

- :warning: Other: 
- None

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -874,6 +874,8 @@ browzine.primo = (function() {
   function shouldAvoidUnpaywall(response) {
     if (response.hasOwnProperty('meta') && response.meta.hasOwnProperty('avoidUnpaywall')) {
       return response.meta.avoidUnpaywall;
+    } else if (response.data.avoidUnpaywallPublisherLinks) {
+      return true;
     } else {
       return false;
     };


tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -4125,18 +4143,121 @@ describe("BrowZine Primo Adapter >", function() {
...
+            "title": "The Residents of The Comoros and Sustainable Tourism",
+            "date": "2023-06-02",
+            "authors": "Sarı, Ömer; Meydan Uygur, Selma; Abdourahmane, Ali",
+            "inPress": false,
+            "abandoned": false,
+            "doi": "10.1163/09744061-bja10082",
+            "linkResolverOpenUrl": "",
+            "pmid": "",
+            "openAccess": false,
+            "unpaywallUsable": true,
```
karlbecker, line 4163: _non-blocking PR comment, no action required_

It strikes me as kind of strange that we have an `unpaywallUsable` value that could very reasonably be true in the API, and we also have `avoidUnpaywallPublisherLinks` ... from what I can tell: 

* `unpaywallUsable` set to true means that we are not actively suppressing Unpaywall data
* `avoidUnpaywallPublisherLinks` means we should not make live unpaywall calls and use the publisher links

So I understand why they each exist, I guess it's just a little odd to me that they could both be at odds and they seem very... similarly named 😅

karlbecker (approved):
Looks good to me @Evroble ! Thanks for addressing it.  I have a little comment about the names of some API elements that, at least right now, strike me as odd, but I know those have distinct meanings so I don't think there's anything else to do for now 😊

tests/acceptance/browzine-primo-adapter-test.js
Evroble: Yeah. It is odd. My understanding is that ```unpaywallUsable``` means a library opts to use unpaywall data if an article  isn't available, whereas ```avoidUnpaywallPublisherLinks``` is very specifically saying if we have a bad publisher link from unpaywall then don't use the unpaywalldata. So yes, a bit confusingly named, but they are distinct I think.

And this example is one I pulled from the API call I did while manually testing, so yeah. The way I read it is "use unpaywall...but not if they have bad publisher links"

karlbecker: Thanks for clarifying that @Evroble and @mike-lang ! I edited the comment, now the definition of `avoidUnpaywallPublisherLinks` is correct after today's meeting.

Just some of the complexity of our system and making the user experience smooth for our users 😊

PR #139, opened 2024-11-27
Add problematic journal support - bz 7970

karlbecker:
## Summary - [BZ-7970](https://thirdiron.atlassian.net/browse/BZ-7970)

Add support for identifying articles that belong to problematic journals

## Description

This allows problematic journal data to be displayed in Primo or Summon.

I've added basic tests that are very similar to the tests made for Expressions of Concern.

One test I didn't bother to add: what happens if an article has both an EOC and is in a problematic journal.  If this happens, EOC should take priority, but I didn't specifically add a test... and I'm not sure it matters because one way or another, the Discovery Service Adapter should display something out of the ordinary beyond it being a typical article.  Think that's OK?

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-7970]: https://thirdiron.atlassian.net/browse/BZ-7970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -579,7 +579,7 @@ browzine.primo = (function() {
       pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
       pdfIconWidth = "15";
       pdfIconMarginRight = "1.5px";
-      articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Problematic Journal";
+      articlePDFDownloadLinkText = browzine.problematicJournalText || "Problematic Journal";
     }
 
     var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
@@ -619,10 +619,10 @@ browzine.primo = (function() {
       articleLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
       articleLinkUrl = problematicJournalArticleNoticeUrl;
-      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
-      pdfIconWidth = "15";
-      pdfIconMarginRight = "1.5px";
-      articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Problematic Journal";
+      linkIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      linkIconWidth = "15";
+      linkIconMarginRight = "1.5px";
+      articleLinkText = browzine.problematicJournalText || "Problematic Journal";
     }
 
     var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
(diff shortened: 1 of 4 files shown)


src/primo/browzine-primo-adapter.js
```diff
@@ -559,6 +574,12 @@ browzine.primo = (function() {
       pdfIconWidth = "15";
       pdfIconMarginRight = "1.5px";
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
+    } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
+      directToPDFUrl = problematicJournalArticleNoticeUrl;
+      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      pdfIconWidth = "15";
```
Evroble, line 580: I wonder if there's a better way to dry up this code, since for retractions, EOC, and problematic journals, the pdfIconWidth and pdfIconMarginRight are all the same size...

src/primo/browzine-primo-adapter.js
```diff
@@ -944,6 +1008,11 @@ browzine.primo = (function() {
           libKeyLinkOptimizer.innerHTML += template;
         }
 
+        if (!directToPDFUrl && !articleLinkUrl && !articleRetractionUrl && !articleEocNoticeUrl && problematicJournalArticleNoticeUrl && isArticle(scope) && showExpressionOfConcern()) {
```
Evroble, line 1011: why do we call "showExpressionOfConcern"? The way i'm reading the code it seems like we're saying "hey we have a problematic journal" so wouldn't we want the "showProblematicJournalUI" function?

src/summon/browzine-summon-adapter.js
```diff
@@ -957,6 +1013,13 @@ browzine.summon = (function() {
           libKeyLinkOptimizer.innerHTML += template;
         }
 
+        if (!directToPDFUrl && !articleLinkUrl && !articleRetractionUrl && !articleEocNoticeUrl && problematicJournalArticleNoticeUrl && isArticle(scope) && showExpressionOfConcern()) {
```
Evroble, line 1016: same note here about "showExpressionOfConcern()"

Evroble (changes requested):
Need further info but otherwise good to go I think

src/primo/browzine-primo-adapter.js
karlbecker: Agreed that it would be nice.  I'd like to focus that on a refactor-only PR in the future if you're cool with that, because we're already adding a hunk of new functionality with this PR.

src/primo/browzine-primo-adapter.js
karlbecker: Good catch @Evroble ! Let me change that and add a test.

I also checked [in Jira](https://thirdiron.atlassian.net/browse/BZ-7970?focusedCommentId=86641) and confirmed that we don't need to have a toggle for turning on or off problematic journals in Discovery Service Adapters.  If that `problematicJournalArticleNoticeUrl` appears in the API, we show it (if there's no retraction or EOC), no other things to check.

karlbecker:
Great review catch @Evroble , we should indeed make that go away!

I checked and I noticed we don't have any automated tests for `articleExpressionOfConcernEnabled`, so I'm not going to add any in this PR.  Could be a future improvement.  But for now I've eliminated the check you noted and the tests are still passing, as I would expect them to, so I think we're in good shape. 

Want to take a look again? Thanks!

Evroble (approved):
LGTM 👍

PR #140, opened 2024-12-10
BZ-9424 SerSol 360 Handles Duplicate ISSNs

Evroble:
## Summary - [BZ-9424](https://thirdiron.atlassian.net/browse/BZ-9424)

A customer reported seeing duplicate links on one entry when searching for a specific journal. We had seen a similar issue before in https://thirdiron.atlassian.net/browse/BZ-5276 and thought we had solved it. This time, the issue has stemmed from duplicate issns being listed, something we did not account for previously. This change adjusts the code to account for duplicate issns and ensure our links are not placed on the same listing multiple times.

## Description

BEFORE:
image

AFTER:
CleanShot 2024-12-10 at 14 58 23

## Implementation Notes

The change largely involves two functions getIssn and getTarget

1). getIssn - this change involves targeting the first issn listed rather than the last, as it seems to be the one more often referenced/used.

2). getTarget - this change involves taking the index from ```titles.forEach``` and using it to pair the appropriate title with the appropriate element so titles[0] attaches to elements[0].

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9424]: https://thirdiron.atlassian.net/browse/BZ-9424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -58,7 +58,7 @@ browzine.serSol360Core = (function() {
   function getTarget(index){
     var elements = $(".results-identifier").closest(".results-title-row");
 
-    if(index > elements.length){
+    if (index >= elements.length) {
       var target = undefined;
     } else {
       var target = elements[index];
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -27,15 +27,15 @@ browzine.serSol360Core = (function() {
 
   function getIssn(title) {
     if (title.identifiers) {
-      var issnIdentifier = title.identifiers.filter(function(identifier) {
+      var issnIdentifierArray = title.identifiers.filter(function(identifier) {
         if (identifier.type && identifier.value) {
           return identifier.type.toLowerCase() === "issn" && identifier.value.length > 0;
         }
-      }).pop();
+      });
 
-      if (issnIdentifier) {
-        return encodeURIComponent(issnIdentifier.value);
-      }
+      if (issnIdentifierArray[0]) {
+        return encodeURIComponent(issnIdentifierArray[0].value);
+      };
 
       var eissnIdentifier = title.identifiers.filter(function(identifier) {
         if (identifier.type && identifier.value) {
@@ -55,15 +55,14 @@ browzine.serSol360Core = (function() {
     return (title && title.title) ? title.title : '';
   };
 
-  function getTarget(title) {
-    var issn = getIssn(title).toLowerCase().trim();
+  function getTarget(index){
+    var elements = $(".results-identifier").closest(".results-title-row");
 
-    var element = $(".results-identifier").filter(function() {
-      return $.trim($(this).text()).toLowerCase().indexOf(issn) > -1 && issn.length > 0;
-    }).closest(".results-title-row");
-
-    var target = element.length > 0 ? element[0] : undefined;
-
+    if(index > elements.length){
+      var target = undefined;
+    } else {
+      var target = elements[index];
+    }
     return target;
   };
 
@@ -81,9 +80,9 @@ browzine.serSol360Core = (function() {
   function addTargets(titles) {
     var titlesToEnhance = [];
 
-    titles.forEach(function(title) {
+    titles.forEach(function(title, index) {
       var issn = getIssn(title);
-      title.target = getTarget(title);
+      title.target = getTarget(index);
       title.endpoint = getEndpoint(issn);
       title.shouldEnhance = shouldEnhance(issn);
 
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -56,7 +56,6 @@ browzine.serSol360Core = (function() {
   };
 
   function getTarget(index) {
-    console.log(index, 'our index inside getTarget');
     var elements = $(".results-identifier").closest(".results-title-row");
 
     if (index >= elements.length) {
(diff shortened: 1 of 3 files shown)


karlbecker:
Hey @Evroble , I think the commit I pushed should get tests passing.

I think `searchResultsCtrl` has to be the name of the element after `$scope`, because that's what this part of the Summon adapter expects:

https://github.com/thirdiron/browzine-discovery-service-adapters/blob/55e3b24177b4a0bdf07c10e2104dd6dbedb36882/src/360-core/browzine-360-core-adapter.js#L100-L114

Let me know if you agree - thanks!

tests/unit/models/360-core.js
```diff
@@ -351,23 +351,6 @@ describe("SerSol 360 Core Model >", function() {
     });
   });
 
-  describe("serSol360Core model getEndpoint method >", function() {
```
karlbecker, line 354: Thanks for axing the duplicate 👍

tests/acceptance/browzine-360-core-adapter-test.js
```diff
@@ -144,5 +144,130 @@ describe("BrowZine SerSol 360 Core Adapter >", function() {
...
+        };
+
+        serSol360Core.adapter(searchResultsSameIssn);
+      });
+
+      /* NOTE to Karl:
+        Essentially what i've tried to do is fully seperate this test from the others, but the code seems to refuse to recognize the data that i've set up
+        I've added logs so you can run locally and see what i'm seeing, but it seems like it's not recognizing the data
+        Previously I tried using 'searchResults' instead of 'searchResultsSameIssn' but then I was getting data from the core test again
+      */
```
karlbecker, line 258: If my changes look good we should be able to remove this comment.

src/360-core/browzine-360-core-adapter.js
```diff
@@ -55,15 +55,15 @@ browzine.serSol360Core = (function() {
...
-    var target = element.length > 0 ? element[0] : undefined;
+  function getTarget(index) {
+    console.log(index, 'our index inside getTarget');
+    var elements = $(".results-identifier").closest(".results-title-row");
 
+    if (index >= elements.length) {
+      var target = undefined;
+    } else {
+      var target = elements[index];
+    }
```
karlbecker, line 66: I think this should work - if you wanted to keep the ternary approach, you could also just do:

```
var target = element.length > index ? element[index] : undefined;
```

Since index is 0-based, and length is 1 based, then it should work fine.

src/360-core/browzine-360-core-adapter.js
```diff
@@ -80,16 +80,18 @@ browzine.serSol360Core = (function() {
 
   function addTargets(titles) {
     var titlesToEnhance = [];
-
-    titles.forEach(function(title) {
+    console.log(titles.length, 'how many titles do we have?');
+    console.log(titles, 'our titles');
```
karlbecker, line 84: 🔥

src/360-core/browzine-360-core-adapter.js
```diff
@@ -80,16 +80,18 @@ browzine.serSol360Core = (function() {
...
       var issn = getIssn(title);
-      title.target = getTarget(title);
+      title.target = getTarget(index);
       title.endpoint = getEndpoint(issn);
       title.shouldEnhance = shouldEnhance(issn);
 
       if (title.shouldEnhance) {
         titlesToEnhance.push(title);
       }
+      console.log(titles, 'our list of titles');
```
karlbecker, line 94: 🔥

src/360-core/browzine-360-core-adapter.js
```diff
@@ -55,15 +55,15 @@ browzine.serSol360Core = (function() {
...
-  function getTarget(title) {
-    var issn = getIssn(title).toLowerCase().trim();
-
-    var element = $(".results-identifier").filter(function() {
-      return $.trim($(this).text()).toLowerCase().indexOf(issn) > -1 && issn.length > 0;
-    }).closest(".results-title-row");
-
-    var target = element.length > 0 ? element[0] : undefined;
+  function getTarget(index) {
+    console.log(index, 'our index inside getTarget');
```
karlbecker, line 59: 🔥

karlbecker (changes requested):
Overall LGTM @Evroble , but we'll need to delete some of these console logs, and probably one of the comments, as long as the test change I made looks good.

Nice job on the sleuthing and getting to the bottom of this mystery! Once those console logs are gone, this is good to merge.

Evroble:
@karlbecker - so I believe i've figured out why my tests won't pass. I left a note in the code but basically in our tests we mock the response through the call ```$.getJSON``` - however in the code, it is expecting only one response, and if there are more than one, it only takes the first. This means I can't really set up the case where multiple titles are getting enhanced, or at least I can't figure out how to set it up in a way to mock that. Once you're back, let's huddle and look at this to see if we can come up with a solution, otherwise, I'm not really sure if there's a good way to test this other than to just ensure that if there are multiple instances of the same issn, we only apply it one time. 

On a separate note, I'd also like us to sit down and maybe add in comments/notes about how the setup of these tests are working, because it is very confusing.

karlbecker:
Works for me @Evroble ! 
In the meantime if you think this fixed the bug, we can move it to testing and ship it, and consider the automated test a good future improvement.

karlbecker (approved):

PR #143, opened 2024-12-17
Refactor code and Update Readme

Evroble:
## Summary - [BZ-9424](https://thirdiron.atlassian.net/browse/BZ-9424)

While working on the bug fix for EJP I noticed a spot where we could refactor the code to make it easier to understand. I also noticed that I never updated the readme to explain the new method of testing primo/summon locally.

This PR does that small code refactor and updates the readme

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9424]: https://thirdiron.atlassian.net/browse/BZ-9424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -287,9 +287,21 @@ Care should be taken to validate JavaScript features used are available in IE11+
...
-* For Primo, add the url given by the ngrok tunnel and add it to the custom.js file of the template package. Then upload that into the School of Mines Primo sandbox. Make sure to change the library ID and API key in order to fetch the data you want if you need to test something like retractions or EOCs.
+* There are two ways to run code locally in order to test changes
+  1. Copy code into a Google override.
+    - When facing a customer problem with Primo for example, this is the easiest way to test changes to their custom.js file. This method is also used when determining problems that may
+      stem from the code, such as when working on SerSol 360 problems.
+      * To do this, open the dev tools in Chrome on the problem instance of Primo. Click the source tab and navigate either to the customer's custom.js file, or to the Amazon S3 file to see our code.
+      * From here you can make changes to the code and press CTRL/CMD + S to save a local override. Note, you may need to enable local overrides by checking the box under the overrides tab ![seen here](<CleanShot 2024-12-17 at 15.30.21.png>)
```
karlbecker, line 295: Thanks for the details and even a screenshot 👍

src/360-core/browzine-360-core-adapter.js
```diff
@@ -214,43 +214,36 @@ browzine.serSol360Core = (function() {
...
-          var boxShadow = "1px 1px 2px #ccc";
-
-          if ($(title.target).find(resultsTitleImageSelector).length > 0) {
-            $(title.target).find(resultsTitleImageSelector).attr("src", coverImageUrl).attr("ng-src", coverImageUrl).css("box-shadow", boxShadow);
-          } else {
-            $(title.target).find(resultsTitleImageContainerSelector).append("<img alt='Results Title Image' class='results-title-image'/>");
-            $(title.target).find(resultsTitleImageSelector).attr("src", coverImageUrl).attr("ng-src", coverImageUrl).css("box-shadow", boxShadow);
+  function searchTitles(titles) {
+
+    titles.forEach(title => {
```
karlbecker, line 219: It looks like IE has supported this since Sep 2012 - https://www.w3schools.com/jsref/jsref_forEach.asp - so I think we're in good shape.  
(MDN's [page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) doesn't even mention IE anymore, heh)

If someone does write in and says things break after this, though, we should be prepared to possibly change from `forEach` to something else, like a regular `for(int i = 0; i < title.length; i++) { ... }` approach.

So I'm going to add a deploy information note here that we should watch out for any reports of this entirely breaking being written in, and if so, change it.  

It shouldn't happen, but juuuust in case.
 
But hooray for getting rid of the recursion, no real need for it here I think 👍

karlbecker (approved):
LGTM @Evroble , thanks for simplifying the code!

PR #147, opened 2025-06-03
remove prefix config prop and fix Article Link logic

danilibros:
## Summary - 

While walking through the current primo code, Karl and I came across some logic that didn't quite seem right. We were only showing an Article Link button if the show direct to PDF link config was also true. Checked with JohnS and verified that we don't need to check the direct to PDF config value when determining if we should display an article link button.


karlbecker (approved):
Phew, thanks for simplifying this logic @danilibros ! LGTM 👍

PR #145, opened 2025-04-17
Doc delivery fulfillment support - bz 9457

karlbecker:
## Summary - [BZ-9457](https://thirdiron.atlassian.net/browse/BZ-9457)

Adds support for document delivery fulfillment

## Description

When the public API has document delivery fulfillment available, Summon and Primo will now use that. 

Primo customization option: `documentDeliveryFulfillmentText`

Summon customization options: `documentDeliveryFulfillmentWording` and `documentDeliveryFulfillmentText`

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9457]: https://thirdiron.atlassian.net/browse/BZ-9457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #149, opened 2025-06-17
Fix when to show doc delivery link for both Primo and Summon - bz 9959

karlbecker:
## Summary - [BZ-9959](https://thirdiron.atlassian.net/browse/BZ-9959)

We should only show the Doc Delivery link if there is no valid article link or direct-to-PDF link.

## Description

When I first made #145 , I thought the public API only put the document delivery fulfillment URL into the API call when that was the only thing that could be used.

After talking with JohnM today, I determined I was wrong! It will show up even if you can access the article, because it's a perfectly valid use case that some folks may want to prefer doc delivery to other URLs in the system.  That's why we show both direct-to-PDF URLs and doc delivery URLs in the same public API response.

@johnmuth , since you haven't reviewed this repo much before, I'd just ask that you check the test describes and its to see if those make sense to you.  If so, then I think the behavior should match what the test description says.

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9959]: https://thirdiron.atlassian.net/browse/BZ-9959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -1106,7 +1106,7 @@ browzine.primo = (function() {
           element.append(template);
         }
 
-        if (browzineWebLink && browzineEnabled && isArticle(scope) && (directToPDFUrl || articleLinkUrl) && showArticleBrowZineWebLinkText()) {
+        if (browzineWebLink && browzineEnabled && isArticle(scope) && (directToPDFUrl || articleLinkUrl || documentDeliveryFulfillmentUrl) && showArticleBrowZineWebLinkText()) {
           var template = browzineWebLinkTemplate(scope, browzineWebLink);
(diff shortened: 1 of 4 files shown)


Evroble (approved):
Remove the console logs/commented out code I noted and this is good to go. I don't fully understand the logic of the document delivery stuff, but seems good as far as I'm following.

johnmuth (approved):
👍 lgtm, fwiw :)

karlbecker:
Thanks for the reviews! Addressed the console logs and mergin' now

PR #150, opened 2025-06-17
Fix Summon button display - bz 9962

karlbecker:
## Description

Remove CSS eq(0) that no longer matches in Summon

Now we use a different CSS selector.

This seemed to have stopped working around June 5 - 7, 2025.

Also corrected a few other classes to ensure this works with the latest version of Summon.

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9962]: https://thirdiron.atlassian.net/browse/BZ-9962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/summon/browzine-summon-adapter.js
+++ b/src/summon/browzine-summon-adapter.js
@@ -1129,7 +1129,7 @@ browzine.summon = (function() {
         }
 
         if (!showLinkResolverLink() && (directToPDFUrl || articleLinkUrl)) {
-          var contentLinkElement = $(documentSummary).find(".availabilityContent");
+          var contentLinkElement = $(documentSummary).find(".availabilityFullText");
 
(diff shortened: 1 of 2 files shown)


karlbecker:
Hey @Evroble , since Summon is completely not working right now, I think a review of this tomorrow would be a very good thing. Perhaps you could try installing it in our sandbox if you've done that before... or have you mostly just done that with Primo?

I have just done it with the override myself, and I think that's giving me enough confidence that it's working fine, so no strict need to do it in the sandbox.  We could have Jay make sure to test it out there.

Thanks for any help you can provide!

PR #154, opened 2025-07-16
BZ-9932-Fix EJP Journal Images on wrong Element

Evroble:
## Summary - [BZ-9932](https://thirdiron.atlassian.net/browse/BZ-9932)

A customer reported that sometimes the journal image for a journal was being incorrectly attached to the wrong journal. 

## Description

While digging into this, my first thought was that something had changed with Syndetics, another service used by libraries to get journal images. But the deeper I dug the more the issues seemed to be off by one. So I looked at how we were finding our targets and that's when I discovered the problem. A few months ago while investigating BZ-9424 I made a change to our target to be based on index instead of an issn, since that was causing an issue with how our view journal in Browzine link was being attached. 

At the time, there wasn't an example of a listing without an identifier, and I wasn't aware that there could be options without an identifier, so I didn't take that into account, which is what caused this problem. In the example library given, one of the listings does not have an identifier, so therefore we skip it, but the indices don't get adjusted and therefore was causing an off by one issue (more if there are more listings without an identifier).

The fix is to just not only consider results with identifiers, our "shouldEnhance" function will deal with that. 

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-9932]: https://thirdiron.atlassian.net/browse/BZ-9932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/360-core/browzine-360-core-adapter.js
+++ b/src/360-core/browzine-360-core-adapter.js
@@ -56,7 +56,7 @@ browzine.serSol360Core = (function() {
   };
 
   function getTarget(index) {
-    var elements = $(".results-identifier").closest(".results-title-row");
+    var elements = $(".results-title-row");
 
     if (index >= elements.length) {
       var target = undefined;


src/360-core/browzine-360-core-adapter.js
```diff
@@ -56,7 +56,7 @@ browzine.serSol360Core = (function() {
   };
 
   function getTarget(index) {
-    var elements = $(".results-identifier").closest(".results-title-row");
+    var elements = $(".results-title-row");
```
karlbecker, line 59: What an elegantly simple bug fix, @Evroble ! 🙌  Glad to see it didn't require a massive amount of change.  

And I think this is one of those changes that we can't really test well with the current EJP test suite, because I think we'd need to be able to test across multiple search results.  You agree?

karlbecker (approved):
LGTM @Evroble 👍

PR #156, opened 2026-01-23
Only use Unpaywall live call responses when "genre" is "journal-article"

karlbecker:
## Summary - [BZ-10184](https://thirdiron.atlassian.net/browse/BZ-10184)

Use the new Unpaywall live logic that is already in other platforms

## Description

Other platforms were able to use the shared Unpaywall library we have, but since this repo doesn't use packages like that, I just added the fairly simple logic into the code directly.

## Implementation Notes

1. ​I let Cursor write the tests, and it seems to have chosen a good test case and a good location for where to insert them... but oh did it like inserting them into the Summon tests.  Like in three places.  Think I should keep them all, or axe one or two sets of those tests?  Seems kind of redundant to me.

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-10184]: https://thirdiron.atlassian.net/browse/BZ-10184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -3674,6 +3791,101 @@ describe("BrowZine Summon Adapter >", function() {
           expect(template.length).toEqual(0);
         });
       });
+
+      describe("unpaywall response without journal-article genre >", function() {
```
Evroble: This section of tests is the same as what's above on line 3282 is it not?? If it's the same, let's get rid of it.

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -4360,6 +4583,110 @@ describe("BrowZine Summon Adapter >", function() {
           expect(quicklink.length).toEqual(1);
         });
       });
+
+      describe("unpaywall response without journal-article genre >", function() {
```
Evroble: Same here? Same set of tests??

Evroble (changes requested):
Overall looks good, but a couple of the tests in the summon adapter test file seem redundant.

tests/acceptance/browzine-summon-adapter-test.js
karlbecker: Yeah, AI did this - that's the Implementation Note I mentioned in the PR.

Since we both had this reaction to it, yeah, let me axe it.

karlbecker:
There you go @Evroble - that look better?

Evroble:
> There you go @Evroble - that look better?

@karlbecker Looks good to me!

Evroble (approved):
LGTM

PR #159, opened 2026-06-22
BZ-10745 Ensure No Duplicate Problematic Journal link because of unpaywall

Evroble:
## Summary - [BZ-10745](https://thirdiron.atlassian.net/browse/BZ-10745)

A customer wrote in about seeing duplicate problematic journal buttons. Karl and I both investigated and came to the same conclusion, that we were doing an Unpaywall call after determining an article was part of a problematic journal, and thus overriding the Unpaywall PDF link with the Problematic journal label, which made it appear like there were two buttons.

## Description

BEFORE:
CleanShot 2026-06-23 at 16 18 43

AFTER:
CleanShot 2026-06-23 at 16 21 39

## Implementation Notes

1. ​Address the problem in the Unpaywall template to return null when we have a problematic journal
2. Add tests to ensure the behavior doesn't regress in future changes.

## Deploy Prerequisites

- None

## Deploy Precautions

- None

## Deploy Informational Notes

- None

[BZ-10745]: https://thirdiron.atlassian.net/browse/BZ-10745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ Care should be taken to validate JavaScript features used are available in IE11+
 * `git clone https://github.com/thirdiron/browzine-discovery-service-adapters.git` this repository
 * change into the new directory
 * `npm install`
+* Note that if you experience failures during the install phase it may be due to ngrok. Try 'npm install --ignore-scripts' and see if that works.
 
 ## Running / Development
 
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -827,7 +827,13 @@ browzine.primo = (function() {
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
       //We don't want to show the PDF link for problematic journal articles as we're already showing the problematic journal article notice
-      return null;
+      directToPDFUrl = problematicJournalArticleNoticeUrl;
+      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      pdfIconWidth = "15";
+      pdfIconMarginRight = "1.5px";
+      articlePDFDownloadLinkText = browzine.problematicJournalText || "Problematic Journal";
+
+      // return null;
     } else if (!directToPDFUrl && showDocumentDeliveryFulfillmentUI(documentDeliveryFulfillmentUrl)) {
       directToPDFUrl = documentDeliveryFulfillmentUrl;
       // pdfIcon stays the same
@@ -891,7 +897,13 @@ browzine.primo = (function() {
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
       //We don't want to show the link for problematic journal articles as we're already showing the problematic journal article notice
-      return null;
+      directToPDFUrl = problematicJournalArticleNoticeUrl;
+      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
+      pdfIconWidth = "15";
+      pdfIconMarginRight = "1.5px";
+      articlePDFDownloadLinkText = browzine.problematicJournalText || "Problematic Journal";
+
+      // return null;
     }
 
     var template = "<div class='browzine' style='line-height: 1.4em; margin-right: 4.5em;'>" +
--- a/src/primo/browzine-primo-adapter.js
+++ b/src/primo/browzine-primo-adapter.js
@@ -827,13 +827,8 @@ browzine.primo = (function() {
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
       //We don't want to show the PDF link for problematic journal articles as we're already showing the problematic journal article notice
-      directToPDFUrl = problematicJournalArticleNoticeUrl;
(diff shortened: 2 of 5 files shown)


karlbecker (commented):
LGTM so far @Evroble , but are there maybe some commits missing? I don't see any tests added to this branch.

Want to add one of those? Then we could merge it in... or you could also add Summon, too, if you had looked at that and determined it needed the same change.

src/primo/browzine-primo-adapter.js
```diff
@@ -826,11 +826,8 @@ browzine.primo = (function() {
       pdfIconMarginRight = "1.5px";
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "Expression of Concern";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
-      directToPDFUrl = problematicJournalArticleNoticeUrl;
-      pdfIcon = "https://assets.thirdiron.com/images/integrations/browzine-retraction-watch-icon.svg";
-      pdfIconWidth = "15";
-      pdfIconMarginRight = "1.5px";
-      articlePDFDownloadLinkText = browzine.problematicJournalText || "Problematic Journal";
+      //We don't want to show the PDF link for problematic journal articles as we're already showing the problematic journal article notice
+      return null;
```
karlbecker: 👍

src/summon/browzine-summon-adapter.js
```diff
@@ -922,10 +922,8 @@ browzine.summon = (function() {
       articlePDFDownloadWording = browzine.articleExpressionOfConcernWording || "Expression of Concern";
       articlePDFDownloadLinkText = browzine.articleExpressionOfConcernText || "More Info";
     } else if (showProblematicJournalArticleNoticeUI(problematicJournalArticleNoticeUrl)) {
-      directToPDFUrl = problematicJournalArticleNoticeUrl;
-      pdfIcon = getRetractionWatchIconSvg();
-      articlePDFDownloadWording = browzine.problematicJournalWording || "Problematic Journal";
-      articlePDFDownloadLinkText = browzine.problematicJournalText || "More Info";
+      //We don't want to show the PDF link for problematic journal articles as we're already showing the problematic journal article notice
```
karlbecker, line 923: Do you think we also want to change the bit on line 854, inside the `unpaywallArticlePDFTemplate` function? Seems like we wouldn't want to show it there either.

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -2179,6 +2179,142 @@ describe("BrowZine Primo Adapter >", function () {
...
+      });
+
+      afterEach(function () {
+        delete browzine.unpaywallEmailAddressKey;
+        delete browzine.articleAcceptedManuscriptPDFViaUnpaywallEnabled;
+
+        jasmine.Ajax.uninstall();
+      });
+
+      it("should show problematic journal notice and not show unpaywall manuscript pdf link", function () {
```
karlbecker, line 2280: This test seems to be right, but I didn't see it committed separately in the commit history, so that we could see it fail in the CI run.

Can you please either make a new branch, or add additional commits to this branch, and confirm that before the logic change in the adapter, these new tests fail, and then after the logic change is brought back, the new tests pass? That's going to make me much more confident about the accuracy of these tests.

tests/acceptance/browzine-summon-adapter-test.js
```diff
@@ -1727,6 +1727,126 @@ describe("BrowZine Summon Adapter >", function() {
...
+      });
+
+      afterEach(function() {
+        delete browzine.unpaywallEmailAddressKey;
+        delete browzine.articleAcceptedManuscriptPDFViaUnpaywallEnabled;
+
+        jasmine.Ajax.uninstall();
+      });
+
+      it("should show problematic journal notice and not show unpaywall manuscript pdf link", function() {
```
karlbecker, line 1813: Same for this too.

tests/acceptance/browzine-primo-adapter-test.js
```diff
@@ -2179,6 +2179,142 @@ describe("BrowZine Primo Adapter >", function () {
       });
     });
 
+    describe("problematic journal notice with unpaywall turned on >", function () {
+      beforeEach(function () {
+        browzine.unpaywallEmailAddressKey = "info@thirdiron.com";
+        browzine.articleAcceptedManuscriptPDFViaUnpaywallEnabled = true;
```
karlbecker, line 2185: Do we also want a test that checks not just when the Unpaywall API responds with a manuscript URL, but also when it just says there's a PDF URL and it's not a manuscript?

karlbecker (changes requested):
Thanks for the ping on the review here, @Evroble ! I'd like to make sure the new tests fail in CI, so we have better confidence that they're truly testing and confirming the behavior change.

I also think I found one more place we'd want to change in Summon, and I think we may want to add another test that isn't just around the manuscript scenario.

Let me know what you think - thanks!

README.md
```diff
@@ -296,6 +296,7 @@ Care should be taken to validate JavaScript features used are available in IE11+
 * `git clone https://github.com/thirdiron/browzine-discovery-service-adapters.git` this repository
 * change into the new directory
 * `npm install`
+* Note that if you experience failures during the install phase it may be due to ngrok. Try 'npm install --ignore-scripts' and see if that works.
```
karlbecker, line 299: I think we don't use grok anymore? Unless you use it or CI relies on it, I think it'd be safe to get rid of it

Evroble: I don't believe we do, but i'm not sure how much effort it would be to remove it, so I felt it was safer to leave that as a separate issue. I'll make a note to write up a Jira that will remove ngrok.

karlbecker (approved):
LGTM @Evroble 👍 thanks for showing the tests were failing in CI, that helped me be much more confident about this change.

